PR #18, opened 2023-02-13
New feature: quiz

yakutovicha:


tutorial/common.py
```diff
@@ -0,0 +1,132 @@
...
+        # Dealing with unanswered questions.
+        not_answered = self.questions_not_answered()
+        if not_answered:
+            self.output.value = f"""<p style="color:Tomato;">Not answerred questions: {", ".join(map(str, not_answered))}!</p>"""
+            return
+
+        # Dealing with erroneous responses.
+        errors = self.erroneous_responses()
+        if errors:
+            self.output.value = f"""<p style="color:Tomato;">Incorrect answers: {", ".join(map(str, errors))}!</p>"""
```
edoardob90, line 124: I would only change the output string when there are incorrect answers. Right now, it says `Incorrect answers: <question number>`. Even a small change like
```html
<p style="color:Tomato;">Your answers to the following questions are <strong>incorrect</strong>: Q1, Q2</p>
```
will be clear enough.

edoardob90 (changes requested):
All good! A small change to output when incorrect answers are given.

baffelli:
My two cents, mostly arising from problems with local development:

- Make sure all dependencies are specified. In this case, I was missing `ipywidgets`. 
- Ideally, change from envrionment.yml to requirements.txt. This is much faster than using conda in most cases. Add that file to this merge request for us to approve it.

edoardob90:
Good point: I was also missing `ipywidgets` and had to install it first

tutorial/common.py
yakutovicha: thanks, fixed in 86f74a0

yakutovicha:
> * Make sure all dependencies are specified. In this case, I was missing `ipywidgets`.

Done in 39976fa, thanks for noticing 👍 

> * Ideally, change from envrionment.yml to requirements.txt. This is much faster than using conda in most cases. Add that file to this merge request for us to approve it.

Honestly, I am not sure this is the right thing to do. Most of the examples [here](https://github.com/binder-examples) use `environment.yml` which offers more flexibilities (also defining dependencies to be [installed through pip](https://github.com/binder-examples/bokeh/blob/master/environment.yml)). The speed is not an issue since the image, once it is built, is not going to be rebuilt again.

If you still think it is important, let's discuss it in a separate issue.

edoardob90 (approved):

PR #20, opened 2023-03-21
Add FAQs

edoardob90:
Adds FAQs for the upcoming tutorial. Might change/be expanded in the future

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -6,7 +6,7 @@ Each tutorial session lasts 4 half-days, from 9:00 to 13:00 (breaks included). D
 
 ## Which operating system should I use?
 
-There is no specific requirement for the operating system. The training will be conducted using Jupyter notebooks, which are browser-based, so an updated browser is the only requirement. Chrome-based or Firefox usually work better with Jupyter notebooks, but this is only a suggestion.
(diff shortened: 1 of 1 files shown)


edoardob90:
A quick comment about the choice of Slack. I figured out that it might be better than Zulip for two reasons:

- I received a few questions like "Where the discussion platform will be hosted? Are we going to use Slack or Teams?"
- There was already a workspace for the tutorial, and I realized that many people already have a Slack account with the `@empa.ch` email address.

I think it's simply the easiest option.

baffelli (dismissed):
Everything ok from my side

FAQ.md
```diff
@@ -0,0 +1,44 @@
+# FAQs for Python Training Course
+
+## What will be the workshop schedule?
+
+Each tutorial session lasts 4 half-days, from 9:00 to 13:00 (breaks included). During one afternoon, a one hour and a half will be dedicated to the extra session on Linux & Gitlab.
+
+## Which operating system should I use?
+
+There is no specific requirement for the operating system. The training will be conducted using Jupyter notebooks, which are browser-based, so an updated browser is the only requirement. Chrome-based or Firefox usually work better with Jupyter notebooks, but this is only a suggestion.
```
yakutovicha, line 9:
```suggestion
There is no specific requirement for the operating system. The training will be conducted on [binder](https://mybinder.org/) using Jupyter notebooks, which are browser-based, so an updated browser is the only requirement. Chrome-based or Firefox usually work better with Jupyter notebooks, but this is only a suggestion.
```

yakutovicha (commented):
all good. Just a minor request.

yakutovicha (approved):

PR #25, opened 2023-03-27
Python environment

baffelli:
Using pure python environment instead of conda.


edoardob90:
Can we close this if we agreed to use Conda + pip environment management?

PR #26, opened 2023-03-27
Control flow

### LINKED ISSUE #7: Control flow
## Basics
- Looping constructs and iterables
- Altering and breaking out of loops: `continue`, `break`
- `if-else` construct and problems with nesting too many branches
- The `else` clause for loops

## Advanced
- The [`match` construct](https://peps.python.org/pep-0636/) (Python 3.10.x) to avoid nesting
- Pros & cons of looping (when it's better to use functional programming methods)
- Recursion and its limits
- The built-in `Exception` class
- Built-in exceptions: `TypeError`, `NameError`, `KeyError`, `ZeroDivisionError`, `OverflowError`
- The `try-except-finally` construct
- How to handle exceptions to have a stack trace useful for debugging

edoardob90:
Fixes #7


control_flow.ipynb
```diff
@@ -0,0 +1,1036 @@
...
+    "    print(contents)\n",
+    "finally:\n",
+    "    file.close()\n",
+    "```\n",
+    "\n",
+    "- We try to open a file called `example.txt`. A `FileNotFoundError` will be raised if it's not found\n",
+    "\n",
+    "- If the file exists, the `else` block is executed: we read and print its contents\n",
+    "\n",
+    "- We use a `finally` block to close the file handle, ensuring that it is properly cleaned up"
```
yakutovicha, line 738: Please add an actual code example here.

yakutovicha (commented):
I just have one suggestion, the rest looks good.

I didn't check the exercises, however.

Once done, go ahead with merging. No need to request my review.

Issue #24, opened 2023-03-27
Add pre-commit hooks


yakutovicha:
Inspiration can be taken from here: https://github.com/nanotech-empa/general-issues/issues/5

PR #27, opened 2023-03-30
Pytest-based testing

### LINKED ISSUE #21: Added branch for testing of solutions
As discussed today with @edoardob90. We are trying to get test suites to run automatically with some cell magic.
Please have a look at the branch testsuite. Unfortunately I am not able to create a merge request because I accidentally pushed these changes to the main branch...let me know if you have a way to fix that other than  my kludges.

edoardob90:
Implement a `pytest` solution for testing. Fixes #21 

Check out the `tutorial/tests/text_example.py` for a template for a generic test class. For each unit of the tutorial, we should create a `test_UNIT.py`. Then, each exercise should have its own `TestExerciseClass` class. It's essential that the `def function_to_test()` is there and marked as a fixture.

In the notebook, the usage is almost identical to `unittest`: `%%pytest_cell exercise_1 test_example::TestClass`. The arguments are the name of the function and the test class.

As `pytest` by default captures all the output of the tests, in case of failure, we're printing **all** the output, including the function's. If we think it's too messy for the users, we can think how to change it.


edoardob90:
>  It's essential that the `def function_to_test()` is there and marked as a fixture.

No, it's not. Last commit is just a small clean-up.

yakutovicha:
hi @edoardob90. I did some changes following your approach. I like it a lot now.

I am sorry for pushing here. My commits can be removed at any time. It just made more sense to implement things directly in the branch instead of giving millions of comments.

I like how simple things have become now 😄

yakutovicha:
This is to explain the idea behind my changes.

To me, it made sense to automate a search for a given function test. Therefore, I propose that tests for a function can always be found under

```python
tutorial/tests/test_{notebook_name}.py::test_{function_to_test}
```

This way, it is enough to put a simple magic call in a cell, and every function in it will be automatically tested 👍 . Of course, we would also need to provide the corresponding tests, but doing so is so simple now.

tutorial/tests/testsuite.py
```diff
@@ -0,0 +1,79 @@
...
+
+@register_cell_magic
+def test_functions_from_cell(line, cell):
+    # Execute the cell code in a new namespace
+
+    # Get the path to the current notebook file
+    nbname = ipynbname.name()
+
+    local_ns = {}
+    exec(cell, local_ns)
```
edoardob90, line 33: This is the only change I would question: it suppresses any IPython output from the cell that's not a side-effect. So, either we force users to **only** define the functions and avoid any other kind of statement, or we must go back to `shell.run_cell(cell)` – but we need a class for that, not simply the `@register_cell_magic` decorator.

yakutovicha: I agree 👍

edoardob90:
All changes discussed with @yakutovicha are done. Please, review and I guess we're good to go.

One important thing: to avoid messing with functions that might be defined in the cell and should **not** be tested, we suggested a naming all test candidates with `solution_FUNCTION_NAME`. In the test module, they should be named `test_FUNCTION_NAME` as usual.

yakutovicha (dismissed):
All good from my side 👍, but I would wait for @baffelli to give a final look.

edoardob90:
Last-minute suggestion: what about naming the test magic just `%%ipytest`? It's much shorter 😁

yakutovicha:
> Last-minute suggestion: what about naming the test magic just `%%ipytest`? It's much shorter 😁

Sounds good to me.

baffelli:
Agree! 👏

baffelli (dismissed):

edoardob90:
Great, we're done! Just the last review for this small change...

yakutovicha:
Looks good, but please remove the output from the notebook cell. Later, I will add a pre-commit hook as a part of https://github.com/empa-scientific-it/python-tutorial/issues/24 to do that automatically.

edoardob90:
> Looks good, but please remove the output from the notebook cell. Later, I will add a pre-commit hook as a part of #24 to do that automatically.

Done ✅

yakutovicha (approved):
perfect 👍 

Just a side note: please make sure to squash-merge the PR to avoid contaminating the commit history.

PR #29, opened 2023-04-06
Replace `lstrip` with string slicing.

### LINKED ISSUE #28: Replace lstrip method with string slicing
It turns out that the`"".lstrip()` method doesn't work as we were expecting. It removes **all** the characters that match the letters provided to the `lstrip()`. Here is the example.

```python
In [2]: "aaaabbbbbccccd".lstrip("cba")
Out[2]: 'd'
```

I think we should replace it with a simple string slicing.

yakutovicha:
fixes #28

### DIFF
--- a/tutorial/tests/testsuite.py
+++ b/tutorial/tests/testsuite.py
@@ -45,7 +45,7 @@ class TestMagic(Magics):
         function_names = {}
         for name, function in self.shell.user_ns.items():
             if name.startswith("solution_") and callable(function):
-                function_names[name[9:]] = function  # Remove the `solution_` prefix
+                function_names[name.removeprefix("solution_")] = function
 
         if not function_names:
             raise ValueError("No function to test defined in the cell")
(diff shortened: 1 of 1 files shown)


tutorial/tests/testsuite.py
```diff
@@ -46,7 +45,7 @@ def ipytest(self, line, cell):
         function_names = {}
         for name, function in self.shell.user_ns.items():
             if name.startswith("solution_") and callable(function):
-                function_names[name.lstrip("solution_")] = function
+                function_names[name[9:]] = function  # Remove the `solution_` prefix
```
edoardob90, line 48: Since we're using Python 3.9+ (hopefully), I think `name.removeprefix('solution_')` is even better than slicing

edoardob90 (changes requested):

tutorial/tests/testsuite.py
yakutovicha: Indeed, it is more explicit 👍

PR #31, opened 2023-04-11
More robust lookup of test module file

### LINKED ISSUE #30: Ipynbname does not work locally
Instead of finding the name of the notebook, `ipynbname.name()` raises a `FileNotFoundError` when working with a local ipython kernel in visual studio code. I suggest using the old way of loading tests where we manually specified the module where they resided. I anyway would prefer a mode explicit solution.

edoardob90:
Fixes #30 

* Handle VS Code behavior
* Optionally specify the module name as the cell magic's argument

### DIFF
--- a/tutorial/tests/testsuite.py
+++ b/tutorial/tests/testsuite.py
@@ -1,12 +1,41 @@
 import io
+import pathlib
 from contextlib import redirect_stdout
 from typing import Callable
 
 import pytest
-import ipynbname
 from IPython.core.interactiveshell import InteractiveShell
 from IPython.core.magic import Magics, magics_class, cell_magic
 from IPython.display import HTML, display
+
+
+def _name_from_line(line: str = None):
+    return line.strip().removesuffix(".py") if line else None
+
+
+def _name_from_ipynbname():
+    try:
+        import ipynbname
+        return ipynbname.name()
+    except FileNotFoundError:
+        return None
+
+
+def _name_from_globals():
+    module_path = globals().get('__vsc_ipynb_file__')
+    if module_path:
+        return pathlib.Path(module_path).stem
+    return None
+
+
+def get_module_name(line: str):
+    """Fetch the test module name"""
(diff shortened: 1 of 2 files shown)


tutorial/tests/testsuite.py
```diff
@@ -1,14 +1,43 @@
...
+
+
+def _name_from_globals():
+    module_path = globals().get('__vsc_ipynb_file__')
+    if module_path:
+        return pathlib.Path(module_path).stem
+    return None
+
+
+def get_module_name(line: str):
```
baffelli, line 31: This is just pedantry, but  if you want to use annotations, add the annotation on the return type

tutorial/tests/testsuite.py
```diff
@@ -33,10 +62,15 @@ class TestMagic(Magics):
     shell: InteractiveShell
 
     @cell_magic
-    def ipytest(self, line, cell):
+    def ipytest(self, line: str, cell: str):
         """The `%%ipytest` cell magic"""
-        # Get the path to the current notebook file
-        notebook_name = ipynbname.name()
+        # Get the module containing the test(s)
+        module_name = get_module_name(line)
```
baffelli, line 68: I like this  solution!

baffelli (changes requested):

tutorial/tests/testsuite.py
```diff
@@ -1,14 +1,43 @@
...
 from IPython.display import HTML, display
 
 
+def _name_from_line(line: str = None):
+    return line.strip().removesuffix(".py") if line else None
+
+
+def _name_from_ipynbname() -> str | None:
+    try:
+        import ipynbname
```
yakutovicha, line 18: Maybe it is better to keep the import at its original place. The module is still present, we are not removing it from the dependencies.

baffelli (dismissed):

yakutovicha (commented):
I like the solution 👍 .

Just a small request, not very important.

yakutovicha (approved):

PR #33, opened 2023-04-13
Fix testsuite and user namespace bug

edoardob90:
Only solution functions defined in the current cell will be tested


yakutovicha (approved):
LGTM!

PR #39, opened 2023-04-15
Fix bug of getting test module name from VS Code

edoardob90:
A small change to `testsuite.py` to fetch the test module name in the right way (hopefully!) from VS Code as well

@baffelli If you can test it/review this, since you're using VS Code 🙏🏻


baffelli:
It works for me.

baffelli (approved):
It works for me.

PR #42, opened 2023-04-16
Add intro slides

edoardob90:
A few introductory slides (max 10 minutes)


intro.ipynb
```diff
@@ -1,99 +1,179 @@
...
+    "%%ipytest\n",
+    "\n",
+    "def solution_to_exercise(input_arg):\n",
+    "    \"\"\"\n",
+    "    Write your solution here\n",
+    "    \"\"\"\n",
+    "    pass\n",
+    "```\n",
+    "\n",
+    "**Unless otherwise noted**, you can delete everything after the line starting with `def` and write your solution.\n",
```
baffelli, line 155: Let's add a reminder that they need to run the cell containing `%load_ext tutorial.tests.testsuite` before the exercises.

baffelli (changes requested):
I suggest adding a note in the slides that I the exercise does not work with the message `UsageError: Cell magic %%ipytest not found` they need to run the cell just before the exercise section where we load the cell magic.

intro.ipynb
```diff
@@ -1,99 +1,179 @@
...
+    "### On Zoom\n",
+    "- Raise your hand 🤚 then ask\n",
+    "- Write in the chat\n",
+    "- We will open a break-out room to address longer questions. **Anyone** interested can jump in\n",
+    "\n",
+    "\n",
+    "### On Slack\n",
+    "- Ask your question in the **corresponding channel**. Use `#help` for generic questions\n",
+    "- Reply **in a thread** if you want to add a comment on another question\n",
+    "- Use `@Sasha`, `@Simone`, `@Edoardo` if you want to mention the tutorial instructors directly\n",
```
yakutovicha, line 105: I think Simone specified their last name too. Another option is to change the name to `Simone` on slack.

```suggestion
    "- Use `@Sasha`, `@Simone Baffelli`, `@Edoardo` if you want to mention the tutorial instructors directly\n",
```

yakutovicha (approved):
Very good!

intro.ipynb
edoardob90: Ok, I'll do that, as discussed

intro.ipynb
baffelli: I changed my name, if you want you can just use `@Simone` now

baffelli (commented):
Great intro, thank you @edoardob90

baffelli (approved):

baffelli (approved):

PR #53, opened 2023-04-17
Fix/show test output

baffelli:
Now showing stdout of tests in any case. Showing the output for each test run separately with cells that can be expanded.


tutorial/tests/testsuite.py
```diff
@@ -49,6 +59,32 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
+        self.tests = set()
+    
+    def pytest_collection_finish(self, session: pytest.Session):
+        [self.tests.add(fun) for fun in session.items]
+
+
+class ResultCollector:
+    """
+    A class that will collect the result of a tes
+    """
```
yakutovicha, line 80:
```suggestion
    """A class that will collect the result of a test."""
```

tutorial/tests/testsuite.py
```diff
@@ -49,6 +59,32 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
 
 
+
+@dataclasses.dataclass
+class TestResult:
+    stdout: str
+    name: str
+
+class TestCollector:
+    """A class to collect all test that will be run"""
```
yakutovicha, line 69:
```suggestion
    """A class to collect all tests that will be run."""
```

tutorial/tests/testsuite.py
```diff
@@ -36,6 +40,12 @@ def get_module_name(line: str, globals_dict: Dict = None) -> str:
 
     return module_name
 
+def find_solution(ns: Dict[str, Callable], postfix: str) -> Optional[Callable]:
+    """
+    Given a namespace, finds a solution matching a given pattern
+    """
```
yakutovicha, line 46:
```suggestion
    """Given a namespace, finds a solution matching a given pattern."""
```

magic_example.ipynb
```diff
@@ -53,12 +52,17 @@
    "execution_count": null,
    "metadata": {},
    "outputs": [],
-   "source": []
+   "source": [
+    "import tutorial.reveal_solution as rs\n",
```
yakutovicha, line 56: This didn't work for me when I was trying.

yakutovicha (changes requested):
Thanks, @baffelli. A few notes:

* I see that you applied code formatted for several files. I think it is a nice thing to do, but I would very much prefer that this is done in a separate PR. Just to simplify reviewers' work.
* I keep thinking that when displaying a solution and the results of the execution, we might organise things using [ipywidgets](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html). Fine to do so in a follow-up PR.

baffelli:
> Thanks, @baffelli. A few notes:
> 
>     * I see that you applied code formatted for several files. This is good but has nothing to do with this PR, so better done in a separate one. Just to simplify reviewers' work. No need to modify this PR, but please keep it in mind for the future pray .
> 
>     * I keep thinking that when displaying a solution and the results of the execution, we might organise things using [ipywidgets](https://ipywidgets.readthedocs.io/en/latest/examples/Widget%20List.html). It is fine to do so in a follow-up PR. But please make an issue then.

Yes I agree. I realised this too late. I will make a separate PR next time.

magic_example.ipynb
baffelli: It was removed in a later commit, let's keep feature branches isolate. Here we only show the output, revealing solutions is for another branch.

edoardob90:
@despadam is working on this. She will open a new one when ready.

yakutovicha:
Shall we delete the PR branch then?

PR #54, opened 2023-04-17
Autoload IPython test magic

edoardob90:
Small caveat: it works **only** from the root path of the notebooks, because the test module is importable from there. Otherwise, we need to tweak the module path(s)


baffelli (approved):
Looks good to me

PR #71, opened 2023-04-20
Bugfix to OOP exercise test

edoardob90:
Fix a bug in the test function of the "N-body problem" exercise. The `function_to_test` was not passed the correct argument


edoardob90:
This also fixes the mistake in the problem's text

yakutovicha (approved):
I think it looks good. Please merge 👍

Issue #77, opened 2023-04-20
Quiz: long strings don't fit in the option RadioButton object


edoardob90:
Look into the docs of `ipywidgets`

PR #78, opened 2023-04-20
Fix quiz on functions

### LINKED ISSUE #74: Functions Quiz Q2: Expected Answer Inconsistent with Tutorial
The tutorial states about type hints:

> Python doesn't require you to explicitly declare a variable's type. However, when defining functions, you can add type hints in the function's signature.
>
> **Type hints improve code readability** and can help catch potential type-related bugs early during development.
>
>The Python interpreter does not enforce type hints and will not check them during runtime. **Type hints are primarily intended for improving code readability**, serving as documentation for developers, and making IDEs much more helpful when writing code. 

However, the question "What is the purpose of type hints in Python function definitions?" is not correctly answered, when the answer "To make the code more readable" is selected.

edoardob90:
Fixes #74

### DIFF
--- a/tutorial/functions.py
+++ b/tutorial/functions.py
@@ -7,7 +7,7 @@ class Functions(Quiz):
             question="How do you indicate a docstring for a Python function?",
             options={
                 "With triple quotes": "Correct!",
-                "With # Docstring:": "",
+                "With # Docstring:": "What's the meaning of # in Python?",
                 "With // Docstring:": "Is // a valid Python comment?",
             },
             correct_answer="With triple quotes",
@@ -17,8 +17,8 @@ class Functions(Quiz):
         q2 = Question(
             question="What is the purpose of type hints in Python function definitions?",
             options={
-                "To enforce strict typing": "",
-                "Only to make the code more readable": "",
+                "To enforce strict typing": "Does the interpreter enforce type hints?",
+                "Only to make the code more readable": "Not only...",
                 "To document the expected types of function arguments and return values": "Correct!",
             },
             correct_answer="To document the expected types of function arguments and return values",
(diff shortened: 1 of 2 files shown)


edoardob90:
This is linked to #77: long options are wrongly formatted otherwise

tutorial/functions.py
```diff
@@ -17,18 +17,18 @@ def __init__(self, title=""):
...
             options={
-                "To enforce strict typing": "",
-                "To make the code more readable": "",
-                "To document the expected types": "Correct!",
+                "To enforce strict typing": "Does the interpreter enforce type hints?",
+                "Only to make the code more readable": "Not only...",
+                "To document the expected types of function arguments and return values": "Correct!",
             },
-            correct_answer="To document the expected types",
+            correct_answer="To document the expected types of function arguments and return values",
```
yakutovicha, line 24: The code readability also must be a part of the correct answer. Otherwise, it should not be mentioned in the advantages of type hints in the text above.

tutorial/functions.py
```diff
@@ -17,18 +17,18 @@ def __init__(self, title=""):
...
             shuffle=True,
         )
 
         q3 = Question(
-            question="What is the syntax to define a function that takes an arbitrary number of arguments in Python?",
+            question="What is the syntax to define a function with an arbitrary number of any type of argument?",
```
yakutovicha, line 29: How about:

Choose a function that accepts an arbitrary number of arguments of any type (positional and keyword).

tutorial/functions.py
yakutovicha: Another option is to rephrase the question so that the answer fits it.

yakutovicha (changes requested):
Thanks, @edoardob90. A few suggestions.

tutorial/functions.py
yakutovicha: sure, we can do it as a follow-up PR.

yakutovicha (approved):
Good 👍

PR #90, opened 2023-04-24
Fix/55

### LINKED ISSUE #55: Mention `enumerate` as a function

baffelli:
Fixes #55


control_flow.ipynb
```diff
@@ -10,24 +10,35 @@
...
+    "    - [The `enumerate` built-in](#The-enumerate-built-in)\n",
+    "    - [The `range` built-in](#The-range-built-in)\n",
+    "  - [Nested loops](#Nested-loops)\n",
+    "  - [Altering loops](#Altering-loops)\n",
+    "    - [`if` statement inside `for`/`while`](#if-statement-inside-for/while)\n",
+    "    - [The `break` keyword](#The-break-keyword)\n",
+    "    - [The `continue` keyword](#The-continue-keyword)\n",
+    "    - [`else` after a `for`/`while`](#else-after-a-for/while)\n",
+    "  - [Exceptions](#Exceptions)\n",
+    "  - [The `try-except` block](#The-try-except-block)\n",
```
edoardob90, line 32:
```suggestion
    "- [References](#References)\n",
    "- [Conditionals](#Conditionals)\n",
    "- [The `while` loop](#The-while-loop)\n",
    "- [The `for` loop](#The-for-loop)\n",
    "    - [The `enumerate` built-in](#The-enumerate-built-in)\n",
    "    - [The `range` built-in](#The-range-built-in)\n",
    "- [Nested loops](#Nested-loops)\n",
    "- [Altering loops](#Altering-loops)\n",
    "    - [`if` statement inside `for`/`while`](#if-statement-inside-for/while)\n",
    "    - [The `break` keyword](#The-break-keyword)\n",
    "    - [The `continue` keyword](#The-continue-keyword)\n",
    "    - [`else` after a `for`/`while`](#else-after-a-for/while)\n",
    "- [Exceptions](#Exceptions)\n",
    "    - [The `try-except` block](#The-try-except-block)\n",
```

Adjust the TOC. No need to include a link to the TOC itself, I think. **Double-check** that it renders correctly.

edoardob90 (changes requested):
Just adjust the TOC itself

baffelli (commented):
The TOC fix worked.

control_flow.ipynb
```diff
@@ -340,6 +349,37 @@
...
+   "outputs": [],
+   "source": []
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5b0747a4",
+   "metadata": {},
+   "source": [
+    "### The `enumerate` built-in\n",
+    "Often,  you need to iterate through a collection, operate on the current value while keeping track of the *position* (index) of the loop.\n",
```
yakutovicha, line 366:
```suggestion
    "Often, you need to iterate through a collection and operate on each value while keeping track of the *position* (index) of the loop.\n",
```

control_flow.ipynb
```diff
@@ -340,6 +349,37 @@
...
+    "In other languages, you would define an index variable and increment it at every loop. While this is possible in python as shown in the previous code snippet, we usually prefer to use the built-in `enumerate` function, documented [here](https://docs.python.org/3/library/functions.html#enumerate).\n",
```
yakutovicha, line 367:
```suggestion
    "In other languages, you would define an index variable and increment it at every loop. While this is possible in Python, as shown in the previous code snippet, we usually prefer to use the built-in `enumerate` function, documented [here](https://docs.python.org/3/library/functions.html#enumerate).\n",
```

yakutovicha (changes requested):
A few very small changes, otherwise good to go 👍

edoardob90 (approved):

yakutovicha (approved):

PR #93, opened 2023-04-24
Update intro slides on exercises

### LINKED ISSUE #46: Explain that all needed inputs are already available
I noticed several people did not seem to understand that the input needed for their solution are already available in the scope of their function. We should write this more explicitely

edoardob90:
Fixes #46 

- [x] Explain input args to solution functions
- [x] Explain how exercise sessions will be organized


intro.ipynb
```diff
@@ -101,7 +102,7 @@
     "\n",
     "### On Slack\n",
     "- Ask your question in the **corresponding channel**. Use `#help` for generic questions\n",
-    "- Reply **in a thread** if you want to add a comment on another question\n",
+    "- Reply **in a thread** if you want to add a comment to another question\n",
```
yakutovicha, line 105:
```suggestion
    "- Reply **in a thread** if you want to add a comment to a question\n",
```
It doesn't matter which question, I believe.

intro.ipynb
```diff
@@ -164,46 +165,73 @@
     "\n",
     "Double-check before deleting everything in a solution's cell. There might be useful hints to solve the exercise!\n",
     "\n",
+    ":::\n",
+    "\n",
+    "---\n",
+    "\n",
+    "Let's look at one important detail about exercises: the variables you are expected to work with in **your solution**.\n",
```
yakutovicha, line 172:
```suggestion
    "Let's look at one important detail about exercises: the variables in **your solution** you are expected to work with.\n",
```

yakutovicha (changes requested):
Thanks, @edoardob90. I added below a few comments.

Also, shouldn't we remove the Quarto slides annotations from this notebook too?

edoardob90:
> Thanks, @edoardob90. I added below a few comments.
> 
> Also, shouldn't we remove the Quarto slides annotations from this notebook too?

Well, I do want to have slides for the intro, not just a notebook: there's no Python code to evaluate. If you want a notebook, than I'll move the slides stuff under `slides/` and leave the notebook without any annotation.

edoardob90:
Done! ✅

baffelli (commented):
Suggested a small change on how to run the cell.

slides/intro_slides.ipynb
```diff
@@ -0,0 +1,271 @@
...
+    "```\n",
+    "\n",
+    "::: {.callout-important}\n",
+    "\n",
+    "The input arguments to the `solution_*` function **is already available**. We also tell you what **type** you should expect: a string, an integer, a file path, or anything else. \n",
+    "\n",
+    ":::\n",
+    "\n",
+    "::: {.notes}\n",
+    "\n",
```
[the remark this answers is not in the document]
edoardob90: 👍🏻 I think we will see that in detail after the intro, since we're doing a "warm up" session on Jupyter only

PR #96, opened 2023-04-25
Remove reload_ext magic from the notebooks

### LINKED ISSUE #76: Remove reload_ext command since the ipytest magic is loaded automatically
Also, make sure that the ipytest indeed works as expected. It can be done by restarting the kernel and executing the test cell right away.

yakutovicha:
fixes #76

### DIFF
--- a/basic_datatypes.ipynb
+++ b/basic_datatypes.ipynb
@@ -333,17 +333,6 @@
    },
    "outputs": [],
    "source": [
-    "%reload_ext tutorial.tests.testsuite"
-   ]
-  },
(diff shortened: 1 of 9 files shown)


edoardob90:
> Does not work on local jupyter, unless you move the script to the folder where your ipython profile is located. We should document this.

Yes, the `ipython_config.py` should be in your local config folder. Or you set the variable `IPYTHONDIR` to `python-tutorial/binder`. I think we should add a comment in the README for those who want to run locally without Docker.

PR #106, opened 2023-05-03
Fix typo in the control flow notebook.

yakutovicha:


edoardob90:
Let's also fix the exercise on calculating the square roots. We should make **super** clear that, if the value doesn't have a defined sqrt, you should skip it.

yakutovicha:
Looks good to me, but I cannot approve it since I created the PR.

edoardob90 (approved):
Good! 😉

PR #108, opened 2023-05-04
Fix tests to I/O exercises

### LINKED ISSUE #109: I/O: fix test to `solution_read_write_file`
This test doesn't seem right:

https://github.com/empa-scientific-it/python-tutorial/blob/da295f8f5d4068299143027b566050cfbcefe98b/tutorial/tests/test_input_output.py#L98-L105

yakutovicha:
fixes #109

### DIFF
--- a/tutorial/tests/test_input_output.py
+++ b/tutorial/tests/test_input_output.py
@@ -30,7 +30,6 @@ def reference_solution_count_parents(f: pl.Path) -> int:
 
 def test_count_parents(function_to_test):
     f = pl.Path(sys.modules[__name__].__file__).parent.parent
-    print(f)
     assert function_to_test(f) == reference_solution_count_parents(f)
 
 
--- a/tutorial/tests/test_input_output.py
+++ b/tutorial/tests/test_input_output.py
@@ -1,45 +1,54 @@
-import pytest
-import pathlib as pl
-import csv 
-import itertools 
-from .testsuite import get_module_name
-from .. import prepare_magic_file
-import sys
-
-from io import StringIO
 import contextlib
+import csv
+import itertools
+import pathlib as pl
+import sys
+from io import StringIO
 
-def get_data(name: str, data_dir:str="data") -> pl.Path:
-    current_module  = sys.modules[__name__]
+import pytest
+
+from .. import prepare_magic_file
+
+
+def get_data(name: str, data_dir: str = "data") -> pl.Path:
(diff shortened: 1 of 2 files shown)


tutorial/tests/test_input_output.py
```diff
@@ -14,7 +14,7 @@ def get_data(name: str, data_dir:str="data") -> pl.Path:
     return (pl.Path(current_module.__file__).parent / f"{data_dir}/{name}").resolve()
 
 def reference_solution_find_all_files(f: pl.Path) -> "list[pl.Path]":
-    return list(f.parent.glob(f.name))
+    return list(f.parent.iterdir())
```
edoardob90, line 17: Also `f.parent.glob("*")` works

yakutovicha: yes, they do the same. So any solution works.

edoardob90:
Also `test_print_salutation` doesn't work. `contextlib` does not have a `redirect_stdin` method. I'm not sure if we want to test that exercise as well.

edoardob90 (changes requested):
Also `test_print_salutation` doesn't work. `contextlib` does not have a `redirect_stdin` method. I'm not sure if we want to test that exercise as well.

yakutovicha:
> > > Also `test_print_salutation` doesn't work. `contextlib` does not have a `redirect_stdin` method. I'm not sure how we want to test that exercise.
> > 
> > 
> > This one is not supposed to be auto-tested. @baffelli wrote it in the description.
> 
> Then let's remove the test, because it's not working and not needed

I let @baffelli comment on this.

baffelli:
> > > > Also `test_print_salutation` doesn't work. `contextlib` does not have a `redirect_stdin` method. I'm not sure how we want to test that exercise.
> > > 
> > > 
> > > This one is not supposed to be auto-tested. @baffelli wrote it in the description.
> > 
> > 
> > Then let's remove the test, because it's not working and not needed
> 
> I let @baffelli comment on this.

Indeed we can remove it. `redirect_stdin` was a silly suggestion from copilot. Anyway I could not find an elegant way to capture stdin, therefore we can safely remove the test.

tutorial/tests/test_input_output.py
baffelli: I did, therefore this is fine. We don't want the solution to be identical :)

PR #114, opened 2023-05-05
Fix exercises on objects.

yakutovicha:


tutorial/tests/test_object_oriented_programming.py
```diff
@@ -21,9 +29,15 @@ def __str__(self):
...
-    assert function_to_test(flavors) == [str(Scoop(flavor)) for flavor in flavors]
+def reference_ice_cream_scoop(flavors: tuple[str]) -> list[Scoop, str]:
+    return [(Scoop(flavor), str(Scoop(flavor))) for flavor in flavors]
+
+
+@pytest.mark.parametrize("flavors", FLAVORS)
+def test_ice_cream_scoop(flavors, function_to_test) -> None:
+    test_solution = [string for _, string in function_to_test(flavors)]
+    reference_solution = [string for _, string in reference_ice_cream_scoop(flavors)]
+    assert test_solution == reference_solution
```
edoardob90, line 40: I've been thinking about this test all afternoon (mostly). And it can be faked easily because it doesn't really enforce you to define/use the class. One can build a tuple with `(None, "<Properly formatted string>")` and the test will pass.

I think it's trickier that it seems to perform some checks on the actual class. One way is to parse the AST as Simone did with some FP tests, but I didn't want to implement it for the time being.

Needs some more thoughts... 💭

baffelli: You could check whether there is a class with the name `Scoop` in the locals() of the function. Something like:

```python
import inspect
import ast

def check_if_class_in_scope(class_name: str, scope: callable) -> bool:
    source = ast.parse(inspect.getsource(scope))
    for node in ast.walk(source):
        match node:
            case ast.ClassDef() as cd:
                return cd.name == class_name


def f1():
    class A:
        a = 1


is_a = check_if_class_in_scope("A", f1)  # True
is_b = check_if_class_in_scope("B", f1)  # True
print(f"A is {is_a}, B is {is_b}")
```

baffelli (changes requested):

tutorial/tests/test_object_oriented_programming.py
edoardob90: That's a good suggestion 👍🏻  We must change the solution and add the class **inside** the solution function. Or apply the same idea to the entire code of the cell.

edoardob90:
Don't know (yet) if we want to fix this issue **before** the next tutorial, but here's something we can do without changing anything substantially: add a test on the **type** (e.g. `isinstance`) of the return value of the solution function instead of only its content.

At the moment, we are only testing the content with `test_solution = [string for _, string in function_to_test(flavors)]`

despadam:
Superseded by https://github.com/empa-scientific-it/python-tutorial/pull/249

Issue #113, opened 2023-05-05
Control flow module, exercise 4


GCatarina:
For clarity, I think the question should say
"4. Write a Python program that takes an integer and divides it **by two** until the result is no longer an even number"

PR #115, opened 2023-05-05
Fix exercises on control flow

### LINKED ISSUE #112: Control flow module, solution of exercise 2 wrong
My solution

```python
def solution_range_of_nums(start: int, end: int) -> list[int]:
    """
    Write your solution here
    """
    if start <= end:
        sol = range(start,end+1)
    else:
        sol = range(start,end-1,-1)
    return [i for i in sol]
```

The accepted solution does not include the last 2 numbers when start > end.
For example, when start = 99, end = 20, accepted solution goes from 99 to 22.

edoardob90:
Fixes #112 and #113

### DIFF
--- a/control_flow.ipynb
+++ b/control_flow.ipynb
@@ -483,7 +483,7 @@
    },
    "outputs": [],
    "source": [
-    "# %%ipytest\n",
+    "%%ipytest\n",
     "\n",
     "def solution_range_of_nums(start: int, end: int) -> list[int]:\n",
     "    \"\"\"\n",
--- a/control_flow.ipynb
+++ b/control_flow.ipynb
@@ -483,7 +483,7 @@
    },
    "outputs": [],
    "source": [
-    "%%ipytest\n",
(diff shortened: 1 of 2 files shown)


control_flow.ipynb
```diff
@@ -483,7 +483,7 @@
    },
    "outputs": [],
    "source": [
-    "%%ipytest\n",
+    "# %%ipytest\n",
```
yakutovicha, line 486: Why did you comment that line? Leftover from testing?

yakutovicha (commented):
I think there is a leftover, the rest looks good 👍

control_flow.ipynb
edoardob90: Yes, I was trying a few quick things and did not want to run the tests. You can make a suggestion to remove it

yakutovicha: Here we go.
```suggestion
    "%%ipytest\n",
```

yakutovicha (approved):

PR #126, opened 2023-07-07
Fix bug and cleanup input/output

### LINKED ISSUE #120: Problem with tests in I/O module?
In the exercises of the I/O module, I am always getting some error message.
Example:

```py
%%ipytest

def solution_print_odd(n: int) -> None: 
    """
    Write your solution here
    """
    for i in range(n+1):
        if i%2==1:
            print(i)
    return None
```

gives the following error in line 94 of tutorial/tests/test_input_output.py:
SyntaxError: unexpected character after line continuation character

My folders are up to date.

edoardob90:
Fixes #120

### DIFF
--- a/tutorial/tests/test_input_output.py
+++ b/tutorial/tests/test_input_output.py
@@ -15,6 +15,27 @@ def get_data(name: str, data_dir: str = "data") -> pl.Path:
     return (pl.Path(current_module.__file__).parent / f"{data_dir}/{name}").resolve()
 
 
+def reference_solution_print_odd(n: int) -> None:
+    for i in range(n):
+        if i % 2 != 0:
+            print(i)
+
+
+@pytest.mark.parametrize("n", [1, 2, 3, 4, 5])
+def test_print_odd(function_to_test, n: int):
+    # redirect stdout of function_to_test to a string
+    solution_stdout = StringIO()
+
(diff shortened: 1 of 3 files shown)


tutorial/tests/test_input_output.py
```diff
@@ -2,17 +2,34 @@
...
 
 import pytest
 
 from .. import prepare_magic_file
 
 
 def get_data(name: str, data_dir: str = "data") -> pl.Path:
-    current_module = sys.modules[__name__]
-    return (pl.Path(current_module.__file__).parent / f"{data_dir}/{name}").resolve()
+    return (pl.Path.cwd() / f"tutorial/tests/{data_dir}/{name}").resolve()
```
baffelli, line 14: I'm not sure about this: the CWD might not correspond to ther location of the module...Otherwise I like this, it's very elegant

baffelli (approved):

tutorial/tests/test_input_output.py
edoardob90: `Path.cwd()` is the directory from where pytest is invoked, that is, `/some/base/path/python-tutorial`. It should work then. I'll double check before merging.

baffelli (approved):

PR #127, opened 2023-07-10
Formatting cell output and catching syntax errors

despadam:
Based on @baffelli 's implementation of `testsuite.py` from branch `fix/show-test-output,` this branch was created from main and contains additional logic in order to:

- Catch syntax errors, using a try-expect block
- Print user output: only print statements written inside the solution functions, all other outputs are disregarded
- Provide readable output when a test fails 

Some Javascript logic was added in order to format the syntax error output. We keep the output as is, but rearrange its position so that it is consistent with the rest of the test result formatting.


tutorial/tests/testsuite.py
```diff
@@ -49,9 +67,136 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
+                    display(HTML(f"<h4>Code output:</h4> {custom_div}{test_outputs[0].stdout}</div>"))
+
+                message = f"<h4>We tested your solution <code>solution_{name}</code> {len(test_outputs)} times with different inputs. "
+                if success:
+                    message += "All tests passed!</h4>"
+                else:
+                    message += "Below you find the details for each test run:</h4>"
+
+                    for test in test_outputs:
+                        test_name = re.search(r'\[.*?\]', test.test_name).group()
```
edoardob90, line 158: Maybe something like:

```py
if match := re.search(r'\[.*?\]', test.test_name):
    test_name = match.group()
    # name formatting
```

tutorial/tests/testsuite.py
```diff
@@ -23,20 +29,32 @@ def _name_from_ipynbname() -> str | None:
...
+        raise RuntimeError(
+            "Test module is undefined. Did you provide an argument to %%ipytest?"
+        )
 
     return module_name
 
 
+def find_solution(ns: Dict[str, Callable], postfix: str) -> Optional[Callable]:
+    """Given a namespace, finds a solution matching a given pattern."""
+    return [f for k, f in ns if k.endswith(postfix)][0]
```
baffelli, line 55: Are you sure it matches all solution patterns?  It only looks for files that end with `postfix`, what happens if we have for example "edoardo_exercise3"? Anyway this function is unused, so we can remove it

tutorial/tests/testsuite.py
```diff
@@ -49,9 +67,136 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
+    Format a long test stdout as a HTML by using the <details> element
+    """
+
+    def join_lines(text: str) -> str:
+        text = text.splitlines()
+        return "".join(
+            [f"<p>{line}</p>" for line in text[-6:-1]]
+        )
+
+    separator = 59 * ' _'
```
baffelli, line 124: Why the fixed value of 59? Is it because the test output is separated by 59 `-`? Can we make it more flexible?

tutorial/tests/testsuite.py
```diff
@@ -49,9 +67,136 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
+        with output_cell:
+            custom_div = '<div style="border: 1px solid; border-color: lightgray; background-color: whitesmoke; margin: 5px; padding: 10px;">'
+            display(HTML("<h3>Test results</h3>"))
+            display(HTML(f"""<div class="alert alert-box {output_config.style}"><h4>{output_config.name}</h4>{output_config.result}</div>"""))
+
+            if not syntax_error:
+                if len(test_outputs) > 0 and test_outputs[0].stdout:
+                    display(HTML(f"<h4>Code output:</h4> {custom_div}{test_outputs[0].stdout}</div>"))
+
+                message = f"<h4>We tested your solution <code>solution_{name}</code> {len(test_outputs)} times with different inputs. "
```
baffelli, line 151: Could you generate this with string interpolation instead of concatenation? I find it easier to read

magic_example.ipynb
```diff
@@ -13,29 +13,139 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 60,
    "metadata": {
     "tags": []
    },
-   "outputs": [],
+   "outputs": [
```
baffelli, line 20: Please remove the outputs from the notebook cells before committing.

baffelli (changes requested):

magic_example.ipynb
despadam: Noted, thanks!

tutorial/tests/testsuite.py
despadam: This was the only unique separator I could find, so that we can keep only the useful output. Everything before the '_ _ _ ...' is not needed and gets ignored. Not an optimal solution, I know, but I can work on it if you'd prefer something more flexible.

tutorial/tests/testsuite.py
despadam: Also added an if statement just in case :)

tutorial/tests/testsuite.py
baffelli: It's fine like this for the time being, thank you!

tutorial/tests/testsuite.py
```diff
@@ -49,9 +59,139 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
+        )
+
+    separator = 59 * ' _'
+    text = text.split(separator)[-1]
+    text_lines = "".join(join_lines(text))
+    test_runs = "".join(
+        [
+            f"""<details style="overflow-y: auto; max-height: 200px;"><summary><u>Click here to expand</u></summary><div>{text_lines}</div></details></li>"""
+        ]
+    )
```
edoardob90, line 123: I find these lines a bit confusing: `join_lines` returns a string, so why on line 118 we are joining again? It doesn't do anything, or am I missing something?

I think all this can be rewritten in a clearer/more compact way. **To be tested**!

```suggestion
    def join_lines(text: str) -> str:
        text = text.splitlines()
        return "".join(
            f"<p>{line}</p>" for line in text[-6:-1]
        )

    separator = 59 * ' _'
    text = text.split(separator)[-1]
    
    test_runs =  f"""<details style="overflow-y: auto; max-height: 200px;"><summary><u>Click here to expand</u></summary><div>{join_lines(text)}</div></details></li>"""
    )
```

tutorial/tests/testsuite.py
```diff
@@ -49,9 +61,167 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
+        return "".join(f"<p>{line}</p>" for line in lines[-6:-1])
+
+    separator = 59 * " _"
+    text = text.split(separator)[-1]
+    text_lines = "".join(join_lines(text))
+    test_runs = "".join(
+        [
+            f"""<details style="overflow-y: auto; max-height: 200px;"><summary><u>Click here to expand</u></summary><div>{text_lines}</div></details></li>"""
+        ]
+    )
```
edoardob90, line 127: I find these lines a bit confusing. A couple of `join` seem unnecessary to me. Joining a string with an empty separator doesn't do anything, right?

My suggestion (**to be tested**!):

```suggestion
   def join_lines(text: str) -> str:
        text = text.splitlines()
        return "".join(
            f"<p>{line}</p>" for line in text[-6:-1]
        )
    separator = 59 * ' _'
    text = text.split(separator)[-1]
    
    test_runs =  f"""<details style="overflow-y: auto; max-height: 200px;"><summary><u>Click here to expand</u></summary><div>{join_lines(text)}</div></details></li>"""
    )
```

baffelli (dismissed):

tutorial/tests/testsuite.py
```diff
@@ -49,9 +61,159 @@ def pytest_generate_tests(self, metafunc: pytest.Metafunc) -> None:
...
+def format_long_stdout(text: str) -> str:
+    """
+    Format a long test stdout as a HTML by using the <details> element
+    """
+
+   stdout_body = re.split(r"_{3,}", text)[-1]
+   html_body = "".join(f"<p>{line}</p>" for line in stdout_body.splitlines()[-6:-1])
+
+   test_runs = f"""<details style="overflow-y: auto; max-height: 200px;"><summary><u>Click here to expand</u></summary><div>{html_body}</div></details></li>"""
+    return test_runs
```
edoardob90, line 120:
```suggestion
def format_long_stdout(text: str) -> str:
    """
    Format a long test stdout as a HTML by using the <details> element
    """
    stdout_body = re.split(r"_{3,}", text)[-1]
    html_body = "".join(f"<p>{line}</p>" for line in stdout_body.splitlines()[-6:-1])

    test_runs = f"""<details style="overflow-y: auto; max-height: 200px;"><summary><u>Click here to expand</u></summary><div>{html_body}</div></details></li>"""

    return test_runs
```

edoardob90 (changes requested):

edoardob90 (approved):

PR #132, opened 2023-07-12
Add pre-commit hooks

### LINKED ISSUE #24: Add pre-commit hooks

yakutovicha:
fixes #24


edoardob90:
Shouldn't `pre-commit` be added as a "dev" dependency somewhere?

yakutovicha:
> Shouldn't `pre-commit` be added as a "dev" dependency somewhere?

At the moment I would add it to `environment.yml`.

If we want to make it a `dev` dependency, we should convert this repo into a proper Python package.

yakutovicha:
Admittedly, I didn't fix all the issues reported by pre-commit. But slowly we will get there 👍

edoardob90 (changes requested):
There might be a few other warnings to add to the ignore list, but I guess this is a good starting point 👍🏻

PR #133, opened 2023-07-28
Reveal reference solutions

despadam:
Reveal reference solution of a test after 3 failed attempts or on success.

Overview:
- There were reference solution function names starting with "reference_" and others starting with "reference_solution_". For consistency reasons, the latter have been renamed so that all start with the prefix "reference_".
- The magic example test file has been modified in order to demonstrate the successful retrieval of all functions used when calling a reference solution function. Feel free to use this for further testing and in the end we can revert it to its original state.
- For readability reasons, testsuite.py has been split into two files and most of the code responsible for the output diplay has been moved to testsuite_helpers.py. The plan is to fully refactor this in a separate branch.

Changes in testsuite.py:
- ipython assigns a unique id to each cell of the notebook. Using this we can count how many times each cell has been executed, so that we only display the proposed solution after 3 failed attempts.
- Using the AST module, we are parsing the file containing the reference solution. We create one Dict for all function defined in the file and one Dict for all imports, where we store their nodes. Then we start parsing the tree of each function that starts with "reference_", creating a Dict where we store the retrieved information for each.
- retrieve_functions() recursively walks the children of each node, to collect information on all functions that are used for a solution. However, the information provided by AST for each node is not sufficient to retrieve the bodies of used functions, hence we need to provide the abovementioned Dicts for defined and imported functions as an argument and retrieve the information from there when needed.
- Finally, when a test is run, we can retrieve its reference solution from the Dict of functions with the "reference_" prefix. We then unprase the nodes to prepare them to be displayed as code. When a used function is imported, we open the file where it comes from an read it from there.

!!! Please note that in order for the last step to be successful, all imports must be written using the absolute path, otherwise AST is unable to locate the file where a function comes from and provides None instead. Example: line 10 in test_input_output.py


tutorial/tests/testsuite.py
```diff
@@ -255,6 +94,39 @@ def ipytest(self, line: str, cell: str):
             if not functions_to_run:
                 raise ValueError("No function to test defined in the cell")
 
+            # store execution count information for each cell
```
baffelli, line 97: I suggest to put the extraction of all definitions and function calls in an external function and call it here. It would make the code more legible

tutorial/tests/testsuite.py
```diff
@@ -277,12 +149,42 @@ def ipytest(self, line: str, cell: str):
...
+                                    solution_code += ast.unparse(node) + "\n\n"
+
                 outputs.append(
                     TestResultOutput(
                         name,
                         False,
                         result == pytest.ExitCode.OK,
                         list(result_collector.tests.values()),
+                        self.cells[cell_id],
+                        solution_code,
```
baffelli, line 187: Could you make Ipython display solution code as a code cell with syntax highlighting?

tutorial/tests/testsuite.py
```diff
@@ -340,10 +242,30 @@ def ipytest(self, line: str, cell: str):
             )
 
 
+def retrieve_functions(
+    all_functions: Dict, node: object, called_functions: Set[str]
+) -> Set[object]:
+    """Recursively walk the AST tree to retrieve all function definitions in a file"""
+
+    for n in ast.walk(node):
+        if isinstance(n, ast.Call) and hasattr(n.func, "id"):
```
baffelli, line 251: You can use a match statement here, for example


```suggestion
match node:
    case ast.Call(ast.Name(id=name), args, keywords):
        """do your work with name here"""
```

```

**Review by baffelli — CHANGES_REQUESTED** [review]

File: tutorial/tests/testsuite.py (hunk 2 above)
Comment (despadam) [reply tutorial/tests/testsuite.py:187]: This is already handled. Please see class TestResultOutput in testsuite_helpers.py

**Comment by yakutovicha** [comment]
Thanks, @despadam, I will have a look today and give my feedback 👍

**Review by yakutovicha — CHANGES_REQUESTED** [review]
Hey @despadam, thanks for the great work!

I finally had time to look at it. Here are some general ideas. Not all of them are directly related to this PR, so it is ok to make an issue instead of fixing them here.

1. The output of the test suite can be rather large, so it is not always obvious where it ends. I suggest drawing a box around it. This will help make it visually more distinct from the rest of the notebook's content.
2. I didn't immediately notice when the solution appeared after a 3rd failed attempt. A message at the bottom that says: "The solution will appear after `x` failed attempts" would help a lot. Finally, you can highlight the solution somehow.
3. This is probably for another PR. When Python fails to compile a function, an exception is raised. We can handle it slightly better.
image
4. When providing solutions to students, one often wants to comment on them. We could parse the solution's docstring and display it to the user. An alternative way is to output the reference solution "as is", including all the comments. This is to discuss and to decide.
5. Often, there is more than one solution. We should support this scenario too. This is maybe for another PR.
6. The output of individual tests is not always super-clear, can we improve it a bit? In the picture below I would:
  - Remove `E assert False`
  - Remove `E +` from each subsequent line
  - Rename`solution_circle_area` with `users_solution`
  - Rename `reference_circle_area ` with `reference_solution`
image

File: tutorial/tests/test_magic_example.py [hunk 4]
```diff
@@ -1,27 +1,59 @@
...
+
+def my_print(text: str) -> None:
+    print(my_test(text))
+
+
+def my_calc(num: int, power: int) -> int:
     return num**power
 
 
+def reference_power(num: int, power: int) -> int:
```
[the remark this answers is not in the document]
despadam: I splitted the function into three separate ones, but I suggest that we keep `test_magic_example.py` and `magic_example.ipynb` for now, as they are used for testing. We can delete them at a later stage.

tutorial/tests/test_functional_programming.py
```diff
@@ -96,20 +96,20 @@ def test_exercise3(
...
     )
 
 
-def reference_solution_exercise4(my_list: List[List[Any]]) -> List[Any]:
+def reference_exercise4(my_list: List[List[Any]]) -> List[Any]:
     return functools.reduce(lambda x, y: x + y, my_list)
 
 
 @pytest.mark.parametrize(
     "my_input, reference_func",
```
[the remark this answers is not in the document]
edoardob90: Yes, definitely it doesn't need to be an argument of `pytest.mark.parametrize`. Can be removed.

tutorial/tests/test_functional_programming.py
```diff
@@ -11,7 +11,7 @@
 from numpy.typing import NDArray
 
 
-def reference_solution_exercise2(my_list: List[int], k: int):
+def reference_exercise2(my_list: List[int], k: int):
```
edoardob90, line 14: This definition should be moved closer to the respective test (above **line 71**). Also:

```suggestion
def reference_exercise2(my_list: List[int], k: int) -> List[int]:
```

tutorial/tests/test_functional_programming.py
edoardob90, line 1: I cannot comment a line not involved in any commit, but [line 79](https://github.com/empa-scientific-it/python-tutorial/blob/80df2c2a4e07dec2530d6296ffcbb712ceac23af/tutorial/tests/test_functional_programming.py#L79) should be changed to

```py
function_to_test: Callable[[List[int, int], int],
```

tutorial/tests/test_input_output.py
```diff
@@ -27,41 +27,41 @@ def test_print_odd(function_to_test, num: int):
...
 
         with contextlib.redirect_stdout(reference_stdout):
-            reference_solution_print_odd(num)
+            reference_print_odd(num)
 
         assert reference_stdout.getvalue() == solution_stdout.getvalue()
 
 
-def reference_solution_find_all_files(f: pl.Path) -> "list[pl.Path]":
+def reference_find_all_files(f: pl.Path) -> "list[pl.Path]":
```
edoardob90, line 35:
```suggestion
def reference_find_all_files(f: pl.Path) -> List[pl.Path]:
```

tutorial/tests/test_input_output.py
```diff
@@ -7,14 +7,14 @@
 
 import pytest
```
edoardob90, line 8:
```suggestion
import pytest
from typing import List, Tuple, Dict
```

tutorial/tests/test_input_output.py
```diff
@@ -27,41 +27,41 @@ def test_print_odd(function_to_test, num: int):
...
 
 
 def test_count_dirs(function_to_test):
     path = pl.Path.cwd()
-    assert function_to_test(path) == reference_solution_count_dirs(path)
+    assert function_to_test(path) == reference_count_dirs(path)
 
 
-def reference_solution_read_file(file: pl.Path) -> "list[str]":
+def reference_read_file(file: pl.Path) -> "list[str]":
```
edoardob90, line 53:
```suggestion
def reference_read_file(file: pl.Path) -> List[str]:
```

tutorial/tests/test_input_output.py
```diff
@@ -92,12 +90,12 @@ def test_read_write_file(function_to_test, tmp_path: pl.Path):
...
 
     function_to_test(input_file, output_file)
-    reference_solution_read_write_file(input_file, test_output_file)
+    reference_read_write_file(input_file, test_output_file)
 
     assert output_file.read_text() == test_output_file.read_text()
 
 
-def reference_solution_exercise1(file: pl.Path) -> dict[str, list[str]]:
+def reference_exercise1(file: pl.Path) -> dict[str, list[str]]:
```
edoardob90, line 98:
```suggestion
def reference_exercise1(file: pl.Path) -> Dict[str, List[str]]:
```

tutorial/tests/test_input_output.py
```diff
@@ -108,37 +106,41 @@ def reference_solution_exercise1(file: pl.Path) -> dict[str, list[str]]:
...
 
 
 def test_exercise2(function_to_test):
     f = get_data("lines.txt")
-    assert function_to_test(f) == reference_solution_exercise2(f)
+    assert function_to_test(f) == reference_exercise2(f)
 
 
-def reference_solution_exercise3(file: pl.Path) -> "dict[str, int]":
+def reference_exercise3(file: pl.Path) -> "dict[str, int]":
```
edoardob90, line 128:
```suggestion
def reference_exercise3(file: pl.Path) -> Dict[str, int]:
```

tutorial/tests/test_input_output.py
```diff
@@ -108,37 +106,41 @@ def reference_solution_exercise1(file: pl.Path) -> dict[str, list[str]]:
...
 def test_exercise3(function_to_test):
     f = get_data("lines.txt")
-    assert function_to_test(f) == reference_solution_exercise3(f)
+    assert function_to_test(f) == reference_exercise3(f)
 
 
-def reference_solution_exercise4(
-    english: pl.Path, dictionary: pl.Path
-) -> list[tuple[str, str]]:
+def reference_exercise4(english: pl.Path, dictionary: pl.Path) -> list[tuple[str, str]]:
```
edoardob90, line 143:
```suggestion
def reference_exercise4(english: pl.Path, dictionary: pl.Path) -> List[Tuple[str, str]]:
```

edoardob90 (changes requested):
My suggestions are minor details (mostly type hints). I did not focus on refactoring `testsuite.py` or the helpers, for the moment.

Really appreciate @despadam implementation! Great job with the AST parsing to retrieve all the names relevant to a solution 👏🏻

despadam:
Hi all,

The latest commit addresses all comments by @yakutovicha and @edoardob90, apart from @yakutovicha's points (4) and (5), which I believe should be addressed in a separate PR.

Thanks a lot!

yakutovicha (dismissed):
Great work, @despadam, thanks a lot!

tutorial/tests/testsuite.py
```diff
@@ -3,20 +3,26 @@
...
 from contextlib import redirect_stderr, redirect_stdout
-from dataclasses import dataclass
-from typing import Callable, Dict, List
+from typing import Dict
 
 import ipynbname
-import ipywidgets
 import pytest
-from IPython.core.display import HTML, Javascript
+from IPython import get_ipython
```
edoardob90, line 10:
```suggestion
from IPython.core.getipython import get_ipython
```

tutorial/tests/testsuite_helpers.py
```diff
@@ -0,0 +1,371 @@
...
+        """
+    return test_runs
+
+
+class TestResultOutput(ipywidgets.VBox):
+    """Class to display the test results in a structured way"""
+
+    def __init__(
+        self,
+        test_outputs: List[TestResult],
```
edoardob90, line 93:
```suggestion
        test_outputs: List[TestResult] | None,
```

tutorial/tests/testsuite.py
```diff
@@ -253,7 +90,18 @@ def ipytest(self, line: str, cell: str):
                     functions_to_run[name.removeprefix("solution_")] = function
 
             if not functions_to_run:
-                raise ValueError("No function to test defined in the cell")
+                raise FunctionNotFoundError()
+
+            # Store execution count information for each cell
+            cell_id = get_ipython().parent_header["metadata"]["cellId"]
```
edoardob90, line 96:
```suggestion
            if (ipython := get_ipython()) is None:
                raise RuntimeError("Could not get IPython instance")
                
            cell_id = get_ipython().parent_header["metadata"]["cellId"]
```

tutorial/tests/testsuite_helpers.py
```diff
@@ -0,0 +1,371 @@
...
+                                        {format_long_stdout(filters.ansi.ansi2html(test.stderr)) if not test.success else ""}
+                                    </div>
+                                """
+                            )
+                        )
+
+                if not reveal_solution:
+                    display(
+                        HTML(
+                            "<h4>&#128221; A proposed solution will appear after 3 failed attempts.</h4>"
```
edoardob90, line 155: This is my only real change suggested: print the remaining failed attempts before showing the solution

```suggestion
                            f"<h4>&#128221; A proposed solution will appear after {3 - cell_exec_count} more failed attempt{'s' if cell_exec_count < 2 else ''}.</h4>"
```

tutorial/tests/testsuite_helpers.py
```diff
@@ -0,0 +1,371 @@
...
+
+class AstParser:
+    """
+    Helper class for extraction of function definitions and imports.
+    To find all reference solutions:
+    Parse the module file using the AST module and retrieve all function definitions and imports.
+    For each reference solution store the names of all other functions used inside of it.
+    """
+
+    def __init__(self, module_file) -> None:
```
edoardob90, line 288:
```suggestion
    def __init__(self, module_file: pathlib.Path) -> None:
```

tutorial/tests/testsuite_helpers.py
```diff
@@ -0,0 +1,371 @@
...
+        self.module_file = module_file
+        self.function_defs = {}
+        self.function_imports = {}
+        self.called_function_names = {}
+
+        tree = ast.parse(open(self.module_file).read())
```
edoardob90, line 294:
```suggestion
        tree = ast.parse(self.module_file.read_text(encoding="utf-8"))
```

tutorial/tests/testsuite_helpers.py
```diff
@@ -0,0 +1,371 @@
...
+
+        for f in solution_functions:
+            if f in self.function_defs:
+                solution_code += ast.unparse(self.function_defs[f]) + "\n\n"
+            elif f in self.function_imports:
+                function_file = pathlib.Path(
+                    f"{self.function_imports[f].replace('.', '/')}.py"
+                )
+                if function_file.exists():
+                    function_file_tree = ast.parse(open(function_file).read())
```
edoardob90, line 356:
```suggestion
                    function_file_tree = ast.parse(function_file.read_text(encoding="utf-8"))
```

edoardob90 (dismissed):
Thanks again, @despadam 🙌! We might have to do some code clean-up, but the "meat" of this (huge) change to the test suite is very good.

Please, just batch-merge the suggestions if you agree with them.

edoardob90 (approved):
LGTM 👍🏻 

We'll sort out the pre-commit checks with mypy in another PR.

binder/ipython_config.py
```diff
@@ -1,6 +1,6 @@
 # Configuration file for ipython.
 
-c = get_config()  # noqa
+c = get_config()  # type: ignore # noqua
```
edoardob90, line 3:
```suggestion
c = get_config()  # noqua
```

edoardob90 (approved):

edoardob90 (approved):

binder/ipython_config.py
```diff
@@ -1,6 +1,6 @@
 # Configuration file for ipython.
 
-c = get_config()  # noqa
+c = get_config()  # noqua
```
edoardob90, line 3:
```suggestion
c = get_config()  # noqa
```

edoardob90 (dismissed):

edoardob90 (approved):

PR #157, opened 2023-11-20
pytest: select tests using a pattern string

edoardob90:


edoardob90:
We are going to apply this change when fixing the async testing.

PR #162, opened 2023-12-05
scikit-learn library

### LINKED ISSUE #145: scikit-learn

fabioacl:
fixes #145


yakutovicha (commented):
Hey @fabioacl. Thank you for your work! It is very good!

I forgot to mention that we try to stick to the "one sentence per line" rule. Since I was going through the notebook, I took the liberty of putting every sentence to a new line. I hope it's OK for you. We do it because it is helpful for the review, as the reviewer can focus on each sentence independently.

Also, when you highlight the inline code, it is better to use single quotes: "`" rather than the triple ones. I also made those changes.

Finally, I updated the TOC and split the last exercise section into one subsection per exercise.

My change request would be to display the data in the `Data preprocessing` section. You can use Matplotlib or some other way that students can see the changes. Maybe showing the size of the arrays after splitting is good enough. It is up to you how to do that.

PR #175, opened 2023-12-12
Make `threading` notebook examples work.

yakutovicha:


threads.ipynb
```diff
@@ -156,7 +156,7 @@
    "source": [
     "We see that the first task to be started (`First task`) finished before the second one could start. \n",
     "This is the sequential computational model we are used to when we first learn programming. \n",
-    "However, in Python we can introduce **parallelism** by using the [multiprocessing](https://docs.python.org/3/library/multiprocessing.html) module. \n",
+    "However, in Python we can introduce **parallelism** by using the [multiprocess](https://github.com/uqfoundation/multiprocess) module. \n",
```
baffelli, line 159: This is a good solution, however we should inform people that `multiprocess` is not the standard module but just the solution to be used on Ipython.

baffelli: Fixed with a warning in be40b2b.

Issue #134, opened 2023-08-27
I/O module, solution of exercise wrong?


GCatarina:
In the exercise "solution_read_write_file", the accepted solution is counting "\n" as a character, in contrast to the example provided.

More clearly, the following solution passes the test:

```python
def solution_read_write_file(input_file: Path, output_file: Path) -> None:
    """
    Write your solution here
    """
    file_r = open(input_file, "r")
    file_w = open(output_file, "w")
    for line in file_r:
        file_w.write(line.strip("\n") + ", " + str(len(line)) + "\n")
        print(line.strip("\n") + ", " + str(len(line)) + "\n") #print output to check
        #file_w.write(line.strip("\n") + ", " + str(len(line)-line.count("\n")) + "\n")
    file_r.close()
    file_w.close()
    return None
```
    
and its output is

```
this, 5 
file, 5 
has multiple, 13 
lines, 6 
how many, 9 
lines, 6 
does this file, 15 
have?, 5
```

I think the commented line should be the right solution instead.

PR #223, opened 2024-04-19
Fixed tests.

baffelli:
Closing #134.


despadam (changes requested):
The table of contents should be updated, for example it does not contain `exercise 3`.

Will also leave some additional comments for errors I came across when trying to execute the exercises.

despadam:
When executing `solution_print_odd` I get `ValueError: I/O operation on closed file`

despadam:
When executing `solution_print_salutation()` I get no output at all.

despadam:
When executing `solution_write_file()` I get `Failed: Cannot read from inexistent file.`

despadam:
When executing `solution_read_write_file()` I get `FileNotFoundError: [Errno 2] No such file or directory: '...\test_read_write_file_solution_0\\output_file.txt'`

baffelli:
> When executing `solution_print_odd()` I get `ValueError: I/O operation on closed file`

Good catch, solved in [27fbfd8](https://github.com/empa-scientific-it/python-tutorial/pull/223/commits/27fbfd8442ea206b1429a2cbebb27859cdaebb18)

baffelli:
> When executing `solution_print_salutation()` I get no output at all.

There's no test for that exercise because I could find no way to use `input` in iyptest

baffelli:
> `solution_write_file`

I changed the error message in 506ac73

baffelli:
> When executing `solution_read_write_file()` I get `FileNotFoundError: [Errno 2] No such file or directory: '...\test_read_write_file_solution_0\\output_file.txt'`

I was using `pytest.fail` instead of `assert`. This causes a confusing error message which is not properly handled by ipytest.
It is now fixed in f0e83ea

baffelli:
> The table of contents should be updated, for example it does not contain `exercise 3`.
> 
> Will also leave some additional comments for errors I came across when trying to execute the exercises.

Done in [eaac18c](https://github.com/empa-scientific-it/python-tutorial/pull/223/commits/eaac18ca68c9e8a3e7519250f779645f2b429f12)

despadam:
Hi @baffelli, all exercises run smoothly now, thanks! The only remaining topic is the TOC:
1. Due to incorrect indentation, the list appears as text instead of links.
2. Due to the bug we have not fixed, when you have multiple sections with the same title, the script will make all links with the same name point to the same section. Here you have multiple sections named `Exercises`, so whichever link you click, it will always send you to the first section with this name. A quick solution is to rename them and make each of the `Exercises` section have a slightly different title.

baffelli:
> Hi @baffelli, all exercises run smoothly now, thanks! The only remaining topic is the TOC:
> 
>     1. Due to incorrect indentation, the list appears as text instead of links.
> 
>     2. Due to the bug we have not fixed, when you have multiple sections with the same title, the script will make all links with the same name point to the same section. Here you have multiple sections named `Exercises`, so whichever link you click, it will always send you to the first section with this name. A quick solution is to rename them and make each of the `Exercises` section have a slightly different title.

Thank you for the hint. I cannot reproduce these problems, for me the links work correctly.

despadam:
> > Hi @baffelli, all exercises run smoothly now, thanks! The only remaining topic is the TOC:
> > ```
> > 1. Due to incorrect indentation, the list appears as text instead of links.
> > 
> > 2. Due to the bug we have not fixed, when you have multiple sections with the same title, the script will make all links with the same name point to the same section. Here you have multiple sections named `Exercises`, so whichever link you click, it will always send you to the first section with this name. A quick solution is to rename them and make each of the `Exercises` section have a slightly different title.
> > ```
> 
> Thank you for the hint. I cannot reproduce these problems, for me the links work correctly.


No worries. If it's ok then I am taking the liberty to rename the sections so the TOC will work correctly for all

despadam (approved):

baffelli (commented):
Looks good to me

PR #227, opened 2024-05-02
Update intro page and add missing titles

### LINKED ISSUE #226: Update intro page and add missing titles
- [x] Update intro page to match the description of the new structure (i.e. 3 days instead of 4)
- [x] Intro page currently has 2 `table of contents` sections. Remove one of them
- [x] Add missing title at the top of the notebook for:
    - [x] input / output
    - [x] modules and packages
    - [x] functional programming

despadam:
Fixes #226


intro.ipynb
```diff
@@ -93,22 +81,15 @@
...
-    "- We will open a break-out room to address long questions. **Anyone** interested can jump in.\n",
-    "\n",
+    "Remember that interaction between us is **essential**.\n",
+    "We encourage you to actively participate and ask questions.\n",
     "\n",
     "## On Slack\n",
     "- Ask your question in the **corresponding channel**. Use `#help` for generic questions.\n",
     "- Reply **in a thread** if you want to add a comment to a question.\n",
-    "- Use `@Sasha`, `@Simone`, `@Edoardo` if you want to mention the tutorial instructors directly.\n",
+    "- Use `@Sasha`, `@Simone`, `@Despina` if you want to mention the tutorial instructors directly.\n",
```
baffelli, line 90: Do we need to keep the slack info considering that we are on-site?

baffelli (changes requested):

intro.ipynb
```diff
@@ -77,13 +65,13 @@
...
     "\n",
-    "We'll be following the following schedule for each day:\n",
+    "Here is the schedule we'll be following each day:\n",
     "\n",
-    "- 8:45-9:00 — Welcome, warm-up questions\n",
-    "- 9:00-10:50 — Topic 1 (theory, hands-on, Q&A)\n",
-    "- *10:50-11:10 — Break* ☕\n",
-    "- 11:10-13:00 — Topic 2\n",
-    "- 🥪 – Practice – Offline discussions on Slack"
+    "- 9:30-10:00 — Welcome coffee and breakfast\n",
```
yakutovicha, line 70: Maybe let's just keep coffee
```suggestion
    "- 9:30-10:00 — Welcome coffee\n",
```

baffelli (commented):
Looks good to me

yakutovicha (changes requested):
Thanks, @despadam, I would only request to change the content of the cell that starts with "As you can see, the message quite clearly tells you that the solution is wrong." 

It shows the old way how results were displayed, it is not the same anymore.

despadam:
> Thanks, @despadam, I would only request to change the content of the cell that starts with "As you can see, the message quite clearly tells you that the solution is wrong."
> 
> It shows the old way how results were displayed, it is not the same anymore.

You are right @yakutovicha. I removed the image and updated the corresponding text.

yakutovicha (approved):
Ship it!

PR #229, opened 2024-05-03
Fix float comparison solution

### LINKED ISSUE #103: Exercise on floats comparison: bug in reference solution
Lines 67-80 of `tutorial/tests/test_basic_datatypes.py` might be wrong. We should use `math.isclose` and maybe provide better test cases

yakutovicha:
fixes #103

### DIFF
--- a/basic_datatypes.ipynb
+++ b/basic_datatypes.ipynb
@@ -646,6 +646,8 @@
    },
    "outputs": [],
    "source": [
+    "import math\n",
+    "\n",
     "a = 0.1 + 0.2\n",
     "b = 0.3\n",
     "print(f\"Is {a} equal to {b}? {a == b}\")\n",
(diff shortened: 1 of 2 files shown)


baffelli (changes requested):
All good. Do we want to add another hint regarding using `isclose` right in the exercise?

yakutovicha:
> All good. Do we want to add another hint regarding using `isclose` right in the exercise?

There is a whole discussion about that in the cell right before the exercise:

image

baffelli:
> > All good. Do we want to add another hint regarding using `isclose` right in the exercise?
> 
> There is a whole discussion about that in the cell right before the exercise:
> 
> image

I noticed. I asked nonetheless, because I am afraid it will be overseen by some.

baffelli (approved):

PR #230, opened 2024-05-03
Fix different issues in datatypes notebook

### LINKED ISSUE #225: Problems in the datatypes notebook
1. The correct solution for the following exercise doesn't work:

```python
%%ipytest

def solution_lists_are_equal(list1: list, list2: list) -> bool:
    # Your code starts here
    return list1 == list2
    # Your code ends here
```
It fails one test case.

2. This example should fail, but it doesn't in Python 3.11. We should add a note
```python
# This should fail.
2.5 + "1.5"
```

3. Exercises on sets are not checked? When I run the cells, nothing shows up.

yakutovicha:
fixes #225

### DIFF
--- a/basic_datatypes.ipynb
+++ b/basic_datatypes.ipynb
@@ -1336,7 +1336,7 @@
    "source": [
     "%%ipytest\n",
     "\n",
-    "def solution_lists_are_equal_but_not_same(list1: list, list2: list) -> bool:\n",
+    "def solution_lists_are_not_same_but_equal(list1: list, list2: list) -> bool:\n",
     "    # Your code starts here\n",
     "    return\n",
     "    # Your code ends here"
@@ -2203,7 +2203,6 @@
    ]
   },
   {
-   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
@@ -2739,7 +2738,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.13"
+   "version": "3.11.1"
   },
   "vscode": {
    "interpreter": {
--- a/basic_datatypes.ipynb
+++ b/basic_datatypes.ipynb
@@ -347,9 +347,8 @@
     "%%ipytest\n",
     "def solution_addition_multiplication(a: float, b:float, c:float) -> float:\n",
     "    # Your code starts here\n",
-    "    solution =\n",
-    "    # Your code ends here\n",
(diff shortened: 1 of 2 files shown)


despadam:
I get the following error for these exercises:
1. `solution_addition_multiplication()`
2. `solution_circle_area()`
3. `solution_quadratic_equation()`

image

despadam:
I get the following error for these exercises:
1. `solution_dict_return_value()`
2. `solution_dict_return_value_delete()`

image

despadam (changes requested):
I noticed that some exercises are not running for me. Am I doing something wrong? Please, see my comments above.

yakutovicha:
@despadam, I addressed the reported issues in https://github.com/empa-scientific-it/python-tutorial/pull/230/commits/5409bac44467d41929c6d90bbbe49a7ecf83cb17, please have a look

basic_datatypes.ipynb
```diff
@@ -2085,9 +2083,9 @@
    "outputs": [],
    "source": [
     "%%ipytest\n",
-    "import typing\n",
-    "T = typing.Typevar(\"T\")\n",
-    "def solution_dict_return_value(my_dict: dict[typing.Hashable, T], key: typing.Hashable) -> typing.Optional[T]:\n",
+    "from typing import TypeVar, Hashable, Any\n",
+    "\n",
+    "def solution_dict_return_value(my_dict: dict[Hashable, Any], key: Hashable) -> Any:\n",
     "    # Your code starts here\n",
```
baffelli, line 2089: Why did you remove the typevar? Now the type is "too generic"?

baffelli (approved):
It looks ok, but I don't see why we should remove the type variables.

basic_datatypes.ipynb
yakutovicha: I thought the value of a dictionary could be any Python object, which I've done by setting the `Any` type. I do not understand the exact problem here.

The original implementation had a bug that @despadam reported above, so I looked around and this is what I found as the solution.

baffelli: It is correct, but if you use a type variable, then the function has a more specific return type. It does not really matter for the moment, so I suggest to merge anyway

despadam (approved):

basic_datatypes.ipynb
despadam: It also looks ok to me and solves the bugs, so I think that since these exercises are anyway trivial, there is no need to worry about typing now

Issue #239, opened 2024-05-08
Errors found in the `object-oriented` notebook


despadam:

despadam:
more of a general comment, not section-specific: we do not reveal the solution when the syntax is incorrect

despadam:
Will add a few simpler exercises

edoardob90:
> more of a general comment, not section-specific: we do not reveal the solution when the syntax is incorrect

Should we?

despadam:
> > more of a general comment, not section-specific: we do not reveal the solution when the syntax is incorrect
> 
> Should we?

@edoardob90 I think not. A more intuitive message for syntax errors would be very useful in these cases, but I think that providing the full solution at that stage would be counter-productive.

despadam:
> Damn it, I noticed this now while testing AI against the new notebook:
> 
> ```python
> class Person:
>         def __init__(self, first_name, last_name, age):
>             self.first_name = first_name
>             self.last_name = last_name
>             self.age = age
> 
> a = Person("Pippo", "Balestra", 45)
> b = Person("Pippo", "Balestra", 45)
> 
> a == b # Always False, because without __eq__ Python compares objects' id
> ```
> 
> @despadam Do we want to fix it or let it be?

I deliberately do not make the `a == b` comparison when testing for this reason, it is doomed to fail. For example, if you look at `Example 3` and `Exercise 3`, I am only testing that the results of the comparisons are the same. 
If this error is just a result of experimenting then I think it is okay to leave it as is. In the theoretical part about comparisons we explain why this fails. So then they are free to play around and get more answers from the AI tools if they wish.

edoardob90:
Or add:

```python
# Validate __eq__ is implemented
    assert (
        "__eq__" in solution_result_a.__class__.__dict__
    ), "Person class must implement __eq__ method for proper equality comparison"
```

But I'll let you decide since you are covering this material in the tutorial.

despadam:
```
assert hasattr(
        solution_result.__eq__, "__closure__"
    ), "Make sure that the class is properly implementing the __eq__() method."
```
I validate the comparison methods like this for `Exercise 3`, but you are right I don't do it for `Example 4`. I will add it there, but leave the rest of the examples and exercises without this validation. Since the rest of them do not touch the topic of comparison, the users are free to experiment as they wish.

PR #249, opened 2024-10-22
Object-oriented: Add simpler exercises

despadam:
#239

### DIFF
--- a/object_oriented_programming.ipynb
+++ b/object_oriented_programming.ipynb
@@ -167,6 +167,48 @@
   {
    "cell_type": "markdown",
    "id": "13",
+   "metadata": {},
+   "source": [
+    "### Example 1"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "14",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%reload_ext tutorial.tests.testsuite"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "15",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-block alert-warning\">\n",
+    "    Inside the following function, write a <code>class</code> called <strong>Person</strong>. The class should have the attributes <code>first_name</code> and <code>last_name</code>. Then, create an instance of this class, representing any person, and return it.\n",
+    "</div>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "16",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%ipytest\n",
+    "\n",
+    "def solution_oop_person():\n",
+    "    # Write your solution here\n",
+    "    pass"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "17",
    "metadata": {
     "tags": []
    },
@@ -176,7 +218,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "14",
+   "id": "18",
    "metadata": {},
    "source": [
     "`width` and `height` are attributes of the `Rectangle` class. But since they are just values (they are **not** functions), we call them **properties**.\n",
@@ -188,7 +230,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "15",
+   "id": "19",
    "metadata": {},
    "source": [
     "You'll note that we were able to retrieve the `width` and `height` attributes (properties) using a dot notation, where we specify the object we are interested in, then a dot, then the attribute we are interested in."
@@ -196,7 +238,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "16",
+   "id": "20",
    "metadata": {},
    "source": [
     "We can add callable attributes to our class (methods), that will also be referenced using the dot notation.\n",
@@ -207,7 +249,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "17",
+   "id": "21",
    "metadata": {
     "tags": []
    },
@@ -228,7 +270,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "18",
+   "id": "22",
    "metadata": {
     "tags": []
    },
@@ -240,7 +282,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "19",
+   "id": "23",
    "metadata": {
     "tags": []
    },
@@ -251,7 +293,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "20",
+   "id": "24",
    "metadata": {},
    "source": [
     "When we ran the above line of code, our object was `r1`, so when `area` was called, Python called the method `area` in the `Rectangle` class automatically, passing `r1` as the argument to the `self` parameter."
@@ -259,7 +301,34 @@
   },
   {
    "cell_type": "markdown",
-   "id": "21",
+   "id": "25",
+   "metadata": {},
+   "source": [
+    "### Example 2\n",
+    "\n",
+    "<div class=\"alert alert-block alert-warning\">\n",
+    "    Modify your code from <strong>Example 1</strong> so that class <code>Person</code> will now have a <strong>method</strong> called <code>full_name</code> that returns the string: <code>My name is {first_name} {last_name}</code>. Then, create an instance of this class, representing any person, and return it.\n",
+    "</div>"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "26",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%ipytest\n",
+    "import inspect\n",
+    "\n",
+    "def solution_oop_fullname():\n",
+    "    # Write your solution here\n",
+    "    pass"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "27",
    "metadata": {
     "tags": []
    },
@@ -269,7 +338,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "22",
+   "id": "28",
    "metadata": {},
    "source": [
     "Special methods are methods that Python defines automatically. If you define a custom class, then you are responsible of defining the **expected behavior** of these methods. Otherwise, Python will fallback to the default, built-in definition, or it will raise an error if it doesn't know what to do.\n",
@@ -279,7 +348,7 @@
   },
   {
    "cell_type": "markdown",
-   "id": "23",
+   "id": "29",
    "metadata": {
     "tags": []
    },
@@ -292,7 +361,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "id": "24",
+   "id": "30",
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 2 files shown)


despadam:
Also added `docstrings` in every solution function.

tutorial/tests/test_object_oriented_programming.py
```diff
@@ -2,6 +2,202 @@
...
+    ],
+)
+def test_oop_fullname(first_name, last_name, function_to_test):
+    solution_result = function_to_test(first_name, last_name)
+    reference_result = reference_oop_fullname(first_name, last_name)
+
+    validate_oop_person(solution_result)
+    validate_oop_fullname(solution_result)
+    assert solution_result.full_name() == reference_result.full_name()
+
```
Snowwpanda, line 93:
```suggestion

    solution_result.first_name = "new_firstname"
    reference_result.first_name = "new_firstname"
    assert solution_result.full_name() == reference_result.full_name()
```

tutorial/tests/test_object_oriented_programming.py
```diff
@@ -2,6 +2,202 @@
...
+
+@pytest.mark.parametrize(
+    "first_name, last_name",
+    [
+        ("Jane", "Doe"),
+        ("John", "Doe"),
+    ],
+)
+def test_oop_fullname(first_name, last_name, function_to_test):
```
Snowwpanda, line 86: Since the exercise says fullname should return My name is {first_name} {last_name}, I can imagine some people will do this which also passes the test but is a difficult mistake to spot in general so I would test that here, below i made a suggestion.

object_oriented_programming.ipynb
```diff
@@ -251,25 +310,64 @@
...
+    "        last_name: a string used to initialize the Person instance\n",
+    "    Returns:\n",
+    "        - an instance of Person\n",
+    "    \"\"\"\n",
+    "\n",
+    "    # Write your solution here\n",
+    "    pass"
+   ]
+  },
+  {
```
Snowwpanda, line 358: I like the new examples with the person, feels intuitive. I added a comment in the test file later for an additional test.
It might also make sense to remind them how importent the self. prefix is when you are using variables of the class itself.

Snowwpanda (changes requested):
The new examples are great. Much more intuitive. Only one change I would make to check the full_name method.

Then things you didn't actually change but it may make sense to rewrite the old exercises more similar to the new examples, e.g. also return the object for better testing and I didn't like the quiz too much, i thought the answers could have been clearer.

tutorial/tests/test_object_oriented_programming.py
despadam: Ahh, or is it that they will use `first_name` instead of `self.first_name`?

tutorial/tests/test_object_oriented_programming.py
```diff
@@ -2,6 +2,221 @@
...
+
+def validate_oop_person(solution_result):
+    params = list(vars(solution_result))
+    assert (
+        type(solution_result).__name__ == "Person"
+    ), "The class should be named 'Person'."
+    assert len(params) == 2, "The class should have 2 attributes."
+    assert (
+        "first_name" in params and "last_name" in params
+    ), "The class attributes should be 'first_name' and 'last_name'."
```
edoardob90, line 29: This is a bit fragile: `vars` requires an object with the `__dict__` method, and if the user returns a string, then we get a `TypeError` which will confuse them.

My suggestion (although it's more verbose):

```suggestion
def validate_oop_person(solution_result):
   # Verify it's not a primitive type
    assert not isinstance(solution_result, (str, int, float, bool, list, dict, tuple)), "Solution must return a class instance, not a primitive"
    # Verify that's a custom class, not a built-in
    assert type(solution_result).__module__ != "builtins", "Solution must return a class instance, not a built-in type"
    # Now the name
    assert type(solution_result).__name__ == "Person", "The class name should be 'Person'"
    
    # Then check the attributes
    try:
        params = list(vars(solution_result))
    except TypeError:
        raise AssertionError("Solution must be a proper class instance with attributes")
    
    assert len(params) == 2, "The class should have 2 attributes"
    assert "first_name" in params and "last_name" in params, \
        "The class attributes should be 'first_name' and 'last_name'"
```

Ideally, using `inspect` is even better, but I think this is good enough.

tutorial/tests/test_object_oriented_programming.py
```diff
@@ -2,6 +2,221 @@
...
+    validate_oop_person(solution_result)
+    validate_oop_fullname(solution_result)
+    assert (
+        solution_result.full_name() == reference_result.full_name()
+    ), "The full_name() result does not match the template 'My name is {first_name} {last_name}'."
+
+    solution_result.first_name = "Jane"
+    solution_result.last_name = "Smith"
+    reference_result.first_name = "Jane"
+    reference_result.last_name = "Smith"
```
edoardob90, line 98: I don't quite get the purpose of these lines. Is this to "verify" that the return value is indeed a class with that attributes?

edoardob90: If yes, then I don't think it's a good idea to "misuse" a class mutability.

tutorial/tests/test_object_oriented_programming.py
edoardob90: Just for _fun_, here's an alternative with `inspect`. The only benefit is that it would work with inheritance (but it's not the case here).

```python
import inspect

# whatever comes before ...

def validate_oop_person(solution_result):
    # First verify we have an object instance, not a primitive
    assert not isinstance(solution_result, (str, int, float, bool, list, dict, tuple)), "Solution must return a class instance, not a primitive type"
    
    # Verify it's a custom class (not a built-in type)
    assert not type(solution_result).__module__ == 'builtins', "Solution must return a custom class instance, not a built-in type"
    
    # Verify the class name
    assert type(solution_result).__name__ == "Person", "The class should be named 'Person'"
    
    # Get attributes using inspect
    attributes = inspect.getmembers(solution_result, lambda a: not inspect.isroutine(a))
    
    # Filter out unwanted attributes
    attributes = [a[0] for a in attributes if not a[0].startswith('_')]
    
    assert len(attributes) == 2, "The class should have 2 attributes"
    assert "first_name" in attributes and "last_name" in attributes, "The class attributes should be 'first_name' and 'last_name'"
```

tutorial/tests/test_object_oriented_programming.py
despadam: It is indeed a "misuse" of class mutability, but the purpose is to enforce the use of internal class attributes when writing the class method. It also took me a while to figure it out (see https://github.com/empa-scientific-it/python-tutorial/pull/249#discussion_r1836107872 above) but the problem is as follows:
- function `reference_oop_fullname()` has attributes `first_name` and `last_name`
- class `Person` also has attributes `first_name` and `last_name`
- when asking users to return the string `My name is {first_name} {last_name}`, if someone uses the function attributes directly, instead of the class attributes, then the tests will still pass
- by adding these lines in the code we make sure to test that they are using the `self.` attributes

tutorial/tests/test_object_oriented_programming.py
despadam: Thanks for the suggestion @edoardob90, it has been addressed with this commit: 56ba50bb4ac3e18bdf2ffd64b4fb18e48ad55929

tutorial/tests/test_object_oriented_programming.py
edoardob90: This is actually very interesting, because the "wrong" class which uses the solution function's parameters would be a **closure**. They would be using something straight from the advanced tutorial! 😄

edoardob90: You could actually add a check like that instead of creating the "extra" person:

```python
def validate_oop_fullname(solution_result):
    methods = [
        attr
        for attr in dir(solution_result)
        if callable(getattr(solution_result, attr)) and not attr.startswith("__")
    ]
    assert len(methods) == 1, "The class should have 1 method."
    assert "full_name" in methods, "The class method should be called 'full_name'."
    assert solution_result.full_name.__closure__ is None, "full_name method should use class attributes (self.)"
```

And then get rid of the `different_person` line. Very elegant!

despadam: This is awesome, I didn't think of it. Thanks! 🙌 Used it here: 1a058da395e3f17d329edf3efdab9f598a8aee1c

edoardob90 (dismissed):
Apart from addressing Example 2's test, I have a comment about the quiz:

#### Q1: `__str__` and `__repr__`

We should say that "`__str__` *can be used* as `__repr__`" because they are technically independent.

```
Q1: Which statement best describes the relationship between __str__ and __repr__?
1. They are independent methods with no relationship to each other
2. __str__ is used as a fallback when __repr__ is missing
3. __repr__ is used as a fallback when __str__ is missing
```
And (3) is the correct one.

#### Q3: `@property`
I would change the correct one to "`@property` creates attributes that act like methods but can be accessed ~~and assigned~~ as regular attributes."

To be able to "set" property attributes, we must user the `@attribute.setter` decorator as well, otherwise Python raises an `AttributeError`.

Also: why using property restricts the dynamic behavior? We could define or change another attribute on-the-fly:

```python
class Circle:
    def __init__(self, radius):
        self.radius = radius
    
    @property
    def area(self):
        self._area = 3.14 * self.radius * self.radius # dynamic, I would say
        return self._area
```

So perhaps rephrase "dynamic" and don't mention attribute setting because we don't talk about setters in this notebook.

edoardob90:
It seems there a lot to fix in this notebook, but it's just that's a tricky topic. Thanks for all the fixes, addition, changes @despadam!

tutorial/tests/test_object_oriented_programming.py
```diff
@@ -2,6 +2,219 @@
...
+        type(solution_result).__module__ != "builtins"
+    ), "Solution must return an instance of a custom class, not a built-in type."
+    assert (
+        type(solution_result).__name__ == "Person"
+    ), "The class should be named 'Person'."
+    # Check the class attributes
+    try:
+        attrs = list(vars(solution_result))
+    except TypeError as e:
+        raise AssertionError from e
```
edoardob90, line 35: If we `raise from`, pytest will catch the `TypeError` first, which will give the output "Your code could not run because of an error". If we **don't want** this, simply raise an `AssertionError` with a message.

edoardob90: This will happen in the simplest case: if the functions returns `None`. So we should add an assertion to exclude a `None` return value. I know it's dumb, but we want to avoid that exception.

despadam: I was struggling with the pre-commit checks because they did not like custom messages when raising an `AssertionError` and this was the easiest workaround. I will try to figure it out.

despadam:
Hey @Snowwpanda & @edoardob90 , thanks for your detailed reviews so far! I finished all changes I wanted to make, so I now ask you for another (hopefully final) review. 

I deliberately left `Exercise 4` as is and did not add validation in the testing, just because I consider it different than the other exercises. That is why it is the only one marked with a 🌶️ now.  I would like to leave users free to interpret it / develop it as they wish, instead of thoroughly testing it - and hopefully let them have fun 😃

edoardob90 (dismissed):
Finally, we're good to go! 🚀  👏🏻

edoardob90 (approved):
Now is the right one! 💯

PR #251, opened 2024-10-22
Fix `functions` notebook errors

### LINKED ISSUE #237: Errors found in the `functions` notebook

edoardob90:
Fix #237

### DIFF
--- a/tutorial/tests/test_functions.py
+++ b/tutorial/tests/test_functions.py
@@ -240,7 +240,7 @@ def test_calculate_area(length, width, unit, function_to_test):
 
 
 def reference_combine_anything(*args: Any) -> Any:
-    """Reference solution for the summing_anything exercise"""
+    """Reference solution for the combine_anything exercise"""
     if not args:
         return args
 


functions.ipynb
Snowwpanda, line 1: Exercise longest_sequence:

Part 1 already tests part 2 as well, I assume this is because of the naming of the exercises? in anycase an n^2 solution doesn't pass part1

I don't like the solution at all. I would show the solution with sorting first, which runs fine and is much better understandable than the O(n) solution, saving a log factor shouldn't matter for them.
If you really want you can have two solutions for part1 and 2 which show a nice improvement by the use of sets and hashing.

For part 2 I get no feedback when run, i could imagine this is again an issue with the naming of the exercise?

Snowwpanda, line 1: Exercise password_validator2:

Solution contains Counter function which is not explained.

Snowwpanda, line 1: Buckets reorginaziation:

Import necessary libraries (pathlib and ascii_lowercase, ascii_uppercase?) in the exercise cells.

Make the reference_solutions self contained. It should be possible to copy paste the ref solution and pass the tests. This also means the input variable names should be consistent

Snowwpanda (changes requested):
Had a look through the changes, the exercises are still too difficult in my opinion but your work definetly helps. I made some comments about older exercises too where I understand with increased difficulty we can expect more from the user, but still I would reorganise a bit, to avoid too much frustration.

tutorial/tests/test_functions.py
```diff
@@ -69,97 +58,172 @@ def test_greet(
...
+@pytest.mark.parametrize(
+    "length,width,unit",
+    [
+        (2.0, 3.0, "cm"),
+        (2.0, 3.0, "m"),
+        (1.5, 2.0, "cm"),
+        (1.5, 2.0, "m"),
+    ],
+)
+def test_calculate_metric_area(length, width, unit, expected, function_to_test):
```
Snowwpanda, line 159:
```suggestion
def test_calculate_metric_area(length, width, unit, function_to_test):
```

This test didn't compile for me, as far as i can tell the expected is not given here so should be removed

Snowwpanda (changes requested):
I like ths split of exercise two, feels much more natural now. I still don't like the formatting, either change or explicitly tell them how to do it with a hind.
Also part 2 needs some change

tutorial/tests/test_functions.py
edoardob90: That's correct, I'll fix it.

functions.ipynb
```diff
@@ -470,59 +483,164 @@
...
+   "id": "32",
+   "metadata": {},
+   "source": [
+    "### Part 1\n",
     "\n",
-    "Write a Python function called `calculate_area` that takes three parameters: `length` (a float), `width` (a float), and `unit` (a string with a **default** value of `\"cm\"`).\n",
-    "The function should calculate the area of a rectangle based on the given length and width, and return the result **as a string** including the correct, default unit (i.e., `cm^2`).\n",
-    "If the unit parameter is \"m\", the function should convert the length and width from meters to centimeters before calculating the area.\n",
+    "Write a function called `solution_calculate_basic_area` that calculates the area of a rectangle given its length and width (in centimeters).\n",
+    "The result should be returned as a string with **two** decimal digits, followed by \"cm^2\".\n",
```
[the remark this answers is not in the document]
edoardob90: I forgot about the f-string formatting. It should actually be mentioned in the "Strings" section. Anyway: I'm changing with `round()` and tell them to use it

Snowwpanda (dismissed):
This looks good, thanks

despadam (approved):

Issue #232, opened 2024-05-06
Errors found in the `basic_datatypes` notebook


despadam:
- [x] Add `import math` in `solution_circle_area()` in basic_datatypes and explain that `pi` is defined there
- [x] It is not clear that quadratic equation expects two results
- [x] `solution_number_is_lower_than_0_or_greater_than_100` in the test we check for `>=`
- [x] in comparison operators, explain the difference of `==` and `isequal` with the float number problem and how is it repsresented in bits

yakutovicha:
in the `def solution_number_is_even(number: float) -> bool:` the input types do not match. It should be `number:int`

baffelli:
https://docs.python.org/3/library/decimal.html

baffelli:
In the exercise `solution_return_first_and_last_element_from_list`, the annotated return type is `list[float]` however the solution checks for a `tuple[float, float]`. We should make this consistent

baffelli:
`popitem` does not remove an arbitrary element but the last element inserted into a dict.

yakutovicha:
For the record: it takes roughly 3 hours to finish the basic datatypes section, so we might want to give students a bit more time for the next tutorial.

despadam:
From #233: In the basic datatypes notebook, trying to convert a complex number to float raises a TypeError as expected. However, it doesn't print the error to the console unless it is surrounded by a try...except statement.

PR #256, opened 2024-10-30
Update on notebook basic_datatypes

Snowwpanda:
Rewrote some of the text. Added some exercises. Fixed some issues.


basic_datatypes.ipynb
```diff
@@ -2105,6 +2310,15 @@
     "from typing import TypeVar, Hashable, Any\n",
     "\n",
     "def solution_dict_return_delete_value(my_dict: dict[Hashable, Any], key: Hashable) -> Any:\n",
+    "    \"\"\"Deletes and returns the value of a key in a dictionary, if it exists, otherwise returns None\n",
```
baffelli, line 2313: I would write more explicitely "the value associated with a key in a dictionary if the key exists. Otherwise, return none".

tutorial/tests/test_basic_datatypes.py
```diff
@@ -467,5 +472,47 @@ def reference_string_split_lines(my_string: str) -> list[str]:
...
 def test_string_split_lines(function_to_test):
     my_string = "\n".join(STRINGS)
-    assert function_to_test(my_string) == STRINGS
     assert function_to_test(my_string) == reference_string_split_lines(my_string)
+
+
+INT_SETS = [set(), {1, 2, 3}, {3, -1, 0, 4, 42, 1002}]
+
+
+def reference_sets_of_even_and_odd(my_set: set[int]) -> tuple(set[int]):
```
despadam, line 481:
```suggestion
def reference_sets_of_even_and_odd(my_set: set[int]) -> tuple[set[int]]:
```

tutorial/tests/test_basic_datatypes.py
```diff
@@ -467,5 +472,47 @@ def reference_string_split_lines(my_string: str) -> list[str]:
...
+            odd_set.remove(n)
+        else:
+            even_set.remove(n)
+    return even_set, odd_set
+
+
+@pytest.mark.parametrize("my_set", INT_SETS)
+def test_sets_of_even_and_odd(my_set, function_to_test):
+    sol_set = my_set.copy()
+    ref_set = my_set.copy()
```
despadam, line 495: Why do we need to copy here? We are already copying inside the reference solution?

despadam (changes requested):
image

This kept happening to me in all exercises, until I fixed the typo in `test_basic_datatypes.py` line 481 (see suggestion below). Can you double-check? Thanks!

tutorial/tests/test_basic_datatypes.py
Snowwpanda: I am sorry you had to look for that error, I was using vscode for a lot of my testing and somehow missed that.

tutorial/tests/test_basic_datatypes.py
Snowwpanda: Technically i only need two clean copies of myset, so one copy would suffice. But if feels more natural to make two new copies and feed them to each of the solution/reference. It's only important that i dont use the same set for function_to_test and the reference solution becuase functoin_to_test might change the set. then my_set stays untouched and we get clean solution functions for both functions. This is then compatible if we for some reason change the reference solution or run a function multiple times for different tests.

tutorial/tests/test_basic_datatypes.py
```diff
@@ -203,15 +203,17 @@ def test_remove_every_second_element_from_list(my_list, function_to_test):
     )
 
 
-def reference_return_first_and_last_element_from_list(my_list: list) -> tuple:
-    return my_list[0], my_list[-1]
+def reference_return_first_and_last_element_from_list(my_list: list) -> list:
```
despadam, line 206: `solution_return_first_and_last_element_from_list` is still expecting a `tuple[float]` as a result

despadam (changes requested):
Also `popitem` does not remove an arbitrary element but the last element inserted into a `dict`. As seen here: https://github.com/empa-scientific-it/python-tutorial/issues/232#issuecomment-2095809160

despadam:
There was also this comment (https://github.com/empa-scientific-it/python-tutorial/issues/232#issuecomment-2100384157) but now I am struggling to find exactly where this happens

tutorial/tests/test_basic_datatypes.py
Snowwpanda: Hmm i think i want the tuple or not? returning the furst and last element with a comma will give a tuple. I think tuple is more consistent with the return format of the rest of the notebooks, or should we put list? then we should also change the return of the function.

Snowwpanda:
> Also `popitem` does not remove an arbitrary element but the last element inserted into a `dict`. As seen here: [#232 (comment)](https://github.com/empa-scientific-it/python-tutorial/issues/232#issuecomment-2095809160)

ah yes, I will fix this


> There was also this comment ([#232 (comment)](https://github.com/empa-scientific-it/python-tutorial/issues/232#issuecomment-2100384157)) but now I am struggling to find exactly where this happens

This is the complex number to float conversion? I could not verify this, for me typeerrors get displayed
image

tutorial/tests/test_basic_datatypes.py
despadam: Yes, let's keep the tuple and make them consistent

despadam:
> > Also `popitem` does not remove an arbitrary element but the last element inserted into a `dict`. As seen here: [#232 (comment)](https://github.com/empa-scientific-it/python-tutorial/issues/232#issuecomment-2095809160)
> 
> ah yes, I will fix this
> 
> > There was also this comment ([#232 (comment)](https://github.com/empa-scientific-it/python-tutorial/issues/232#issuecomment-2100384157)) but now I am struggling to find exactly where this happens
> 
> This is the complex number to float conversion? I could not verify this, for me typeerrors get displayed image



Hey!
Requesting a cosmetic change as well, as this part appears like this for me:
image

Then, if you scroll a bit down, just above the `Quiz on basic data type arithmetics` section, there is a cell with conversion examples. I believe what is missing there is to also add a cell that does this:
```
# complex -> float
res = float(complex_number)
print(f"complex -> float: Result is {res}. Type of the result {type(res)}")
```
This one will print the expected TypeError, which also for me works without a `try...except` statement.

tutorial/tests/test_basic_datatypes.py
despadam: > Yes, let's keep the tuple and make them consistent

Once we also address this, I think we are good to go 👍

yakutovicha:
I am now on it as well. I will do a quick pass.

despadam (approved):

basic_datatypes.ipynb
```diff
@@ -2491,6 +2841,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "Important to note: when unpacking you will recieve a list instead of whatever iterable you had to begin with. This may be relevant when further processing data.\n",
```
yakutovicha, line 2844: Please keep one sentence per line. Please check the whole "Unpacking" section for this. Many sentences here are written in one line (which makes it harder to review).

yakutovicha (changes requested):
Thanks a lot, @Snowwpanda, great work 👍 .

Here are some comments/suggestions:

1. I always have problems reading this sentence: "However, the returned float is rounded down to the nearest integer, which is 2.0 in the example above.” Can it be somehow nicely rephrased?
2. Please remove the inputs from the cell at the top and place them where relevant:
```python
import sys
import math
from tutorial.quiz import basic_datatypes as bdt
```
3. Section “Unpacking” has some sentences written in one line. Please keep them one sentence per line.

basic_datatypes.ipynb
```diff
@@ -321,8 +322,8 @@
     "In the cell below you should write the code that solves the following exercises:\n",
     "\n",
     "  1. Write a Python program to that gets three float numbers and retuns $(a + b) * c$.\n",
-    "  2. Write a Python program that computes the area of a circle with radius $r$.\n",
-    "  2. Write a Python program that returns a solution of quadratic equation $ax^2 + bx + c = 0$. We consider only the case when solutions are real numbers."
+    "  2. Write a Python program that computes the area of a circle with radius $r$. Use math.pi from the math library as value of pi.\n",
+    "  2. Write a Python program that returns the solutions of the quadratic equation $ax^2 + bx + c = 0$. We consider only the case when there are two solutions and both are real numbers."
```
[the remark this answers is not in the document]
Snowwpanda: No the "two" solutions can be the same and then I wouldn't consider this valid, that's what I meant by two. 
E.g. I don't allow x^2 -1 = 0 because then it's unclear, should I return 1 or return (1,1)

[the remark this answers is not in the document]
Snowwpanda: I will write distinct, i think that makes it clearer

basic_datatypes.ipynb
```diff
@@ -303,7 +290,7 @@
    "source": [
     "<div class=\"alert alert-block alert-info\">\n",
     "<b>Note:</b>: The floor division operation `//` returns a `float` in case of floats.\n",
```
edoardob90, line 292: I still find this a bit verbose. My suggestion:
```suggestion
    "<b>Note:</b> The floor division operator <code>//</code> performs division and returns the largest integer less than or equal to the result.\n",
```

basic_datatypes.ipynb
```diff
@@ -303,7 +290,7 @@
...
-    "However, the returned float is rounded down to the nearest integer, which is 2.0 in the example above.\n",
+    "However, the returned float is floored to the nearest lower integer, which is -3.0 in the example above.\n",
```
edoardob90, line 293: Linked to the previous suggestion:
```suggestion
    "When operating on floats, it returns a float result, like -3.0 in the example above. When operating on integers, it returns an integer result.\n",
```

basic_datatypes.ipynb
```diff
@@ -633,7 +668,10 @@
     "math.isclose(a, b)\n",
     "```\n",
     "\n",
-    "Try it for yourself in the cell below."
+    "Try it for yourself in the cell below.\n",
+    "\n",
+    "Note: The math.isclose will have a default of 1e-09 deviation tolerance.\n",
```
edoardob90, line 673:
```suggestion
    "**Note:** The `math.isclose` will have a default of $1\\times 10^{-9}$ deviation tolerance.\n",
```

edoardob90 (changes requested):
You can apply all of these in batch.

edoardob90 (approved):
I fixed a few more paragraphs to adhere to the "one sentence per line" rule. But otherwise, I think that's finally good to go. Thanks @Snowwpanda 👏🏻

PR #258, opened 2024-11-11
Fix errors found in the `input_output` notebook

Snowwpanda:
Updated notebook input_output


tutorial/tests/test_input_output.py
```diff
@@ -86,15 +107,18 @@ def test_write_file(function_to_test, tmp_path: pl.Path):
...
-        # Here we iterate over each line of the input file
-        for line in read_file.readlines():
-            # We remove line breaks and write the line to the output file
-            clean_line = line.strip("\n\r")
-            # We crete the output line
-            output_line = f"{clean_line}, {len(clean_line)}\n"
-            # Finally we write the line and its length to the output file
-            write_file.write(output_line)
+    read_file = input_file.open("r")
+    write_file = output_file.open("w")
```
baffelli, line 111: Why did you remove the context managers? I think it is a good practice to introduce them early.

input_output.ipynb
```diff
@@ -193,7 +193,7 @@
     "    <h4><b>Important</b></h4>\n",
     "    ⚠️ Note that <code>input</code> is a <b>blocking operation</b>. The program evaluation stops and waits until the user provides an input. Therefore, use <code>input</code> <b>carefully</b> and only when truly necessary. A common issue is that someone first writes an interactive program expecting user input through <code>input</code> and later integrates it into a larger application that is supposed to run automatically without any user interaction. Suddenly, the application stops somewhere and does not run further because of a well hidden <code>input</code> call...<br><br>\n",
```
despadam, line 194:
```suggestion
    "    ⚠️ Note that <code>input</code> is a <b>blocking operation</b>. The program execution  stops and waits until the user provides an input. Therefore, use <code>input</code> <b>carefully</b> and only when truly necessary. A common issue is that someone first writes an interactive program expecting user input through <code>input</code> and later integrates it into a larger application that is supposed to run automatically without any user interaction. Suddenly, the application stops somewhere and does not run further because of a well hidden <code>input</code> call...<br><br>\n",
```

input_output.ipynb
```diff
@@ -465,7 +500,15 @@
...
-    "op.Paths()"
+    "op.Paths()\n",
+    "from pathlib import Path\n",
+    "\n",
+    "\n",
+    "current_path = Path(\"tutorial/tests/data/\")\n",
+    "print(list(current_path.iterdir()))\n",
+    "#Output all file names in that directory\n",
+    "for file in current_path.iterdir():\n",
+    "    print(file.name)"
```
despadam, line 511: Why is this code here? When executing the cell, the quiz does not appear

input_output.ipynb
```diff
@@ -1074,9 +1170,15 @@
...
+    "for i in range(0, 6):\n",
+    "    bt = i.to_bytes(4)\n",
+    "    print(f\"bytes: {[hex(b) for b in bt]}, integer: {i}, string: {chr(i)}\")\n",
+    "for i in range(60, 74):\n",
+    "    bt = i.to_bytes(4)\n",
+    "    print(f\"bytes: {[hex(b) for b in bt]}, integer: {i}, string: {chr(i)}\")\n",
+    "for i in range(8**3, 8**3 + 3):\n",
     "    bt = i.to_bytes(4)\n",
-    "    print(f\"bytes: {bt}, integer: {i}, string: {chr(i)}\")"
+    "    print(f\"bytes: {[hex(b) for b in bt]}, integer: {i}, string: {chr(i)}\")"
```
despadam, line 1181: This cell didn't print any output for me

despadam (changes requested):
Addressed a few details in the notebook

tutorial/tests/test_input_output.py
Snowwpanda: I agree that they are good practice, but it isn't nice to have them in the solutions before properly introduced in the tutorial so here i write it the old fashion way.

input_output.ipynb
Snowwpanda: Python version issue, thanks for catching that

input_output.ipynb
Snowwpanda: My mistake

tutorial/tests/test_input_output.py
```diff
@@ -86,15 +107,18 @@ def test_write_file(function_to_test, tmp_path: pl.Path):
...
+        clean_line = line.strip("\n\r")
+        # We crete the output line
+        output_line = f"{clean_line}, {len(clean_line)}\n"
+        # Finally we write the line and its length to the output file
+        write_file.write(output_line)
+    read_file.close()
+    write_file.close()
 
 
 def test_read_write_file(function_to_test, tmp_path: pl.Path):
```
despadam, line 124: When I tried testing this, I wrote only the strings (without removing the line endings and without writing the string length) but it would anyway pass. Then I saw that the file `lines.txt`, which is used for testing, is empty.

input_output.ipynb
```diff
@@ -1074,9 +1162,15 @@
...
+    "    print(f\"bytes: {[hex(b) for b in bt]}, integer: {i}, string: {chr(i)}\")\n",
+    "for i in range(60, 74):\n",
+    "    bt = i.to_bytes(4, 'little')\n",
+    "    print(f\"bytes: {[hex(b) for b in bt]}, integer: {i}, string: {chr(i)}\")\n",
+    "for i in range(8**3, 8**3 + 3):\n",
+    "    bt = i.to_bytes(4, 'little')\n",
+    "    print(f\"bytes: {[hex(b) for b in bt]}, integer: {i}, string: {chr(i)}\")"
    ]
   },
   {
```
despadam, line 1176: The cell below, under `Converting bytes to text`, when executed give me `TypeError: to_bytes() missing required argument 'byteorder' (pos 2)`

despadam (changes requested):
A couple more comments

tutorial/tests/test_input_output.py
Snowwpanda: I couldn't find a place where we explicitly change the lines.txt, however it is used both in test exercises as well as the explanation part and is never reset which makes it easy to change the file accidently. I took a new file now for the tests and try to reset for exercises after the write introduction.

input_output.ipynb
Snowwpanda: Hmm don't think i ever changed that..for consitency i will add the byteorder

PR #259, opened 2024-11-12
Run tutorial in GitHub Codespaces

edoardob90:
Adds a minimal `devcontainer.json` to be able to run the tutorial inside GitHub Codespaces.

This is another alternative to Binder.


Snowwpanda (commented):
This is nice. 
I tried it out and worked perfectly with the jupyter editor. 
It didn't work in the VS Code editor, but this is an issue i have locally as well to some degree, i think its a problem with threads blocking eachother?
A bit more explanations for beginners might be needed in the readme if we want to offer this as a "feature".
I tried a couple of notebooks and couldn't find anything missing, but maybe someone else should have a look too.

despadam:
Very good idea! It also works well for me when running it in JupyterLab. VS Code does run into some issues. 
The only downside in my case was that building the codespace was a bit slow, but I still find it a good alternative.

edoardob90:
> Very good idea! It also works well for me when running it in JupyterLab. VS Code does run into some issues. The only downside in my case was that building the codespace was a bit slow, but I still find it a good alternative.

There's a way to have "pre-built" codespaces, but I didn't read carefully that part of the docs.

despadam (approved):

edoardob90:
Okay, configuring prebuilds is not complicated (although I had to add a payment method because we _may_ have to pay a few cents for storing the builds), but there's a tricky situation to handle: the codespace container depends on the repository Docker image which is automatically built after each successful push to `main`. If the codespace prebuild is triggered **before** the new image is pushed to the registry, then it's useless. Maybe GitHub is clever enough, but I'm not betting on it.

I thought of creating another, manual GitHub Actions, but again I'm not sure if it's the right way. Here it is:

```yaml
name: Prebuild Codespace

on:
  workflow_run:
    workflows: ["Build Tutorial Container"]
    types:
      - completed
    branches:
      - main

jobs:
  prebuild:
    if: ${{ github.event.workflow_run.conclusion == 'success' }}
    runs-on: ubuntu-latest
    permissions:
      contents: read

    steps:
      - name: Codespace prebuild
        run: |
          gh codespace create \
          --repo ${{ github.repository }} \
          --branch main \
          --location WestEurope \
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

despadam:
Perhaps a scheduler would be a good back up plan, for not only relying on rebuilding when pushing to main.

edoardob90:
**Note:** We'll leave this PR open until early next week. We can then merge it and configure the prebuilds to optimize starting up Codespaces during the tutorial.

despadam:
Before merging this, we should also update the `intro` page with relevant instructions. For now we only have a section called `Binder only`.

Issue #167, opened 2023-12-11
Errors found in `manage_python_project.ipynb` notebook


yakutovicha:
Please report here any errors or  suggestions for the [`manage_python_project.ipynb`](https://github.com/empa-scientific-it/python-tutorial/blob/main/manage_python_project.ipynb).

despadam:
Typo in **Preparatory exercise**:
- **other** yours or the one you have forked --> should be 'either'

fabioacl:
In the line `bumpver update --minor --dry-run` it should be `--dry` and not `--dry-run`

fabioacl:
In the Testing section, add `job` before `tests`.

anates:
Executing `bumpver update --minor --dry-run` as proposed fails, as there is no entry for `__version__` in `__init__.py` in `mypackage`. Manually adding this entry resolves the issue.

edoardob90:
There seems to be an issue on Windows and the `yamlfmt` hook. I don't know if it's reproducible every time, but it seems to be a bug in the hook script itself. Check out: https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/issues/25

yakutovicha:
The exercise on releasing automation needs to be revised. It is not clear to people what is `test.pypi.og` and how to release things there. Maybe, it is enough to release only on GitHub.

yakutovicha:
> Last entry in the notebook for publishing a new release, currently reading
> 
> ```
> - uses: softprops/action-gh-release@v0.1.14
>    name: Create release
> ```
> 
> should be
> 
> ```
>     - name: Create release
>       uses: softprops/action-gh-release@v0.1.14  
> ```
> 
> afaik. Is that correct?

It should not matter, as after reading it becomes a dictionary:
```python
{
    "name": "Create release",
    "uses": "softprops/action-gh-release@v0.1.14",
}
```

but I agree this way is more consistent. I will fix it.

yakutovicha:
> Executing `bumpver update --minor --dry-run` as proposed fails, as there is no entry for `__version__` in `__init__.py` in `mypackage`. Manually adding this entry resolves the issue.

This I will do directly in the `mypackage` repo.

PR #261, opened 2024-11-14
Pre-commit: add Ruff

### LINKED ISSUE #255: Add Ruff as a pre-commit hook?
A suggestion: why not replacing `flake8` and others pre-commit hooks (like `isort`, `pyupgrade`) with Ruff, which can do almost everything? Including a slightly better formatting than Black.

Resources:
- https://docs.astral.sh/ruff/linter/#ruff-check
- https://github.com/astral-sh/ruff-pre-commit

edoardob90:
Fixes #255 

- Replaces flake8 and all its enabled plugins with Ruff
- Use Ruff formatter as a drop-in replacement of Black
- Simplifies the configuration using a single `pyproject.toml` instead of multiple files (`pytest.ini`, `flake8`).


edoardob90:
There are a few errors to fix in the code (or to add to the ignore, while it's not the ideal way):

- [ ] `TRY003`: Avoid specifying long messages outside the exception class
- [ ] `E741`: Ambiguous variable name

Snowwpanda:
Thanks for starting this, since a change on flake8 pre-commit is not passing, so why not add ruff in one go. I fixed the places where pre-commit was not passing, don't understand enough about the rest though, so someone else should approve or I will read later.

pyproject.toml
```diff
@@ -0,0 +1,50 @@
...
+    "TRY301", # Raise within try block
+    # "W503"  # Line break before binary operator (not PEP8 enforced, so not implemented in Ruff)
+]
+
+# flake8 plugins to enable:
+# - flake8-bugbear          B
+# - flake8-builtins         A
+# - flake8-comprehensions   C4
+# - flake8-debugger         T10
+# - flake8-logging-format   G
```
despadam, line 22: I am also not the most fit to review this, but I do have a question. Why is `G` not included in the list below?

edoardob90: Because I simply forgot it! 😅 I did it manually, no Copilot. Good catch! 🔎

edoardob90:
I think this is good to go.

Just a comment on why we ignore [`B904`](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/) from the object-oriented tests: we do want to raise an assertion error if the check on the solution function attributes fails, but Python by default will raise a `TypeError` via `vars(solution_result)` if the result is not a class and therefore doesn't have a `__dict__` attribute.

despadam:
> I think this is good to go.
> 
> Just a comment on why we ignore [`B904`](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/) from the object-oriented tests: we do want to raise an assertion error if the check on the solution function attributes fails, but Python by default will raise a `TypeError` via `vars(solution_result)` if the result is not a class and therefore doesn't have a `__dict__` attribute.

Thanks for the explanation! Can we document this in a comment?

despadam:
I tried testing it and this is what happens in my case:

image

edoardob90:
> I tried testing it and this is what happens in my case:
> 
> image

Seems okay to me. It's the "annoying" yaml formatter that keeps replacing 2 spaces with 4. Many editors are okay with 2. It happens to me as well. We could actually fix this...

edoardob90:
I think the problem with yaml formatting is fixed now.

edoardob90:
> > I think this is good to go.
> > Just a comment on why we ignore [`B904`](https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/) from the object-oriented tests: we do want to raise an assertion error if the check on the solution function attributes fails, but Python by default will raise a `TypeError` via `vars(solution_result)` if the result is not a class and therefore doesn't have a `__dict__` attribute.
> 
> Thanks for the explanation! Can we document this in a comment?

Good idea: I added it in 9bdfe5d

despadam:
> I think the problem with yaml formatting is fixed now.

I keep having problems with `YAML (check)` and `YAML (format)`. Check always complains about `wrong new line character: expected \n`, even after I fix it. Format always fails. Maybe I am doing something wrong? We can have a look together during tomorrow's meeting.

Snowwpanda:
> I tried testing it and this is what happens in my case:
> 
> image

Ah and here i thought this was just me, after seeing you have the same issue, this seems to be a known problem:
https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/issues/32
where windows path is messed up, i ran on codespaces (linux env) where it worked fine. but since some of us are using windows this isn't so nice.

Snowwpanda:
I could fix the YAML formating using the pull request fix they have. Instead of 
rev 0.2.3 
i used     
rev: 71af8556bfcf12745fdfdb3ed57b6698b444f58c
See here for the PR:
https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/pull/42
And the discussion:
https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/issues/25#issuecomment-1466943391
Not sure if this is a nice fix though, but i think its that or leave the hook out entirely

edoardob90:
> I could fix the YAML formating using the pull request fix they have. Instead of rev 0.2.3 i used rev: 71af8556bfcf12745fdfdb3ed57b6698b444f58c See here for the PR: [jumanjihouse/pre-commit-hook-yamlfmt#42](https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/pull/42) And the discussion: [jumanjihouse/pre-commit-hook-yamlfmt#25 (comment)](https://github.com/jumanjihouse/pre-commit-hook-yamlfmt/issues/25#issuecomment-1466943391) Not sure if this is a nice fix though, but i think its that or leave the hook out entirely

My memory didn't help me here: I already raised this problem [last year](https://github.com/empa-scientific-it/python-tutorial/issues/167#issuecomment-1850143839). There were indeed a few users running the tutorial locally, and that means 99% of the times on Windows.

edoardob90:
Okay, I found an alternative formatter that _should_ work on Windows as well. Please, Windows people, double check. I know it's marginally relevant since most people will use Codespaces/Binder/whatever.

despadam:
Yaml format works for me now, but Yaml lint still fails.

edoardob90:
@despadam @Snowwpanda As discussed, I remove `yamllint`. Everything works, including the formatting. Also, we have already a basic `check-yaml` hook built-in in pre-commit. I think it's enough for our purposes.

despadam (approved):
Works as expected now, thanks!

.pre-commit-config.yaml
```diff
@@ -1,68 +1,38 @@
...
+  - repo: https://github.com/google/yamlfmt
+    rev: v0.14.0
+    hooks:
+      - id: yamlfmt
+        name: YAML (format)
+        types: [yaml]
+  - repo: https://github.com/asottile/setup-cfg-fmt
+    rev: v2.7.0
+    hooks:
+      - id: setup-cfg-fmt
```
yakutovicha, line 34: Maybe delete this one too? I am not sure we are using it at all.

edoardob90: Indeed. We don't have `setup.cfg` files. I'll go ahead, remove that and merge this.

PR #264, opened 2024-11-20
Simplify and optimize Docker image

edoardob90:
- Create a simpler Docker image starting from official [Jupyter Docker Stacks](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html)
- Remove `jupyter-repo2docker` dependency
- Add GitHub workflow to build & push the image (as before)
- Adjust the Conda environment which is configured from `docker/environment.yml`

The old workflow has been renamed to `.github/workflows/repo2docker.yml` and Binder is still supported via the `binder/` directory (unchanged).

Also: since Docker provides already the isolation, there's no need to have another virtual environment. All the tutorial dependencies are installed in the `base` environment, which is automatically configured by the base image.


yakutovicha:
@edoardob90 do you need a review on this? I could give a hand.

edoardob90:
> @edoardob90 do you need a review on this? I could give a hand.

If you want, you can try to build the container from the Dockerfile. I tested the workflow in the `main` branch of [my fork](https://github.com/edoardob90/python-tutorial), and everything seems to work fine. But a double-check is always appreciated 😉

Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+    fix-permissions "${CONDA_DIR}" && \
+    fix-permissions "/home/${NB_USER}"
+
+# Prepare IPython configuration (move earlier in the build)
+RUN mkdir -p ${HOME}/.ipython/profile_default
+COPY --chown=${NB_UID}:${NB_GID} binder/ipython_config.py ${HOME}/.ipython/profile_default/
+
+# Copy the repository late in the build process
+RUN mkdir -p ${REPO}
+COPY --chown=${NB_UID}:${NB_GID} . ${REPO}/
```
yakutovicha, line 45: This part is a bit tricky. The user should either start the **same** container (so the readme needs to be updated), or we should mount the folder to the container.

In any case, the readme should be synchronized with the new approach.

docker/post-build.sh
yakutovicha, line 1: Is this file used anywhere?

docker/setup_custom_env.py
yakutovicha, line 1: Also not sure why this file is needed at all.

yakutovicha (changes requested):
Thanks, @edoardob90, I did a first pass and I would need some clarifications from your side.

docker/post-build.sh
edoardob90: Nope, I added those steps to the Dockerfile

docker/setup_custom_env.py
edoardob90: Same as above: I added those steps to the Dockerfile

edoardob90: No, actually this is just there "for reference". I'm not creating an environment inside Docker, which seems redundant.

Dockerfile
edoardob90: I think the step is correct and makes sense only when we're building the image. The user can still mount the local folder with a local copy of the repository, and nothing changes if the are no mismatches. Otherwise they will have their local repository inside the container.

edoardob90: It might be good to create a docker volume, but I have to think about it carefully.

edoardob90: I did a bit of research. It seems that those lines **are not needed**: I added them to make sure to have the repository material inside the container, but:

1. If the user is mounting a local folder with the repository, that's redundant
2. GitHub Codespaces, which can use a prebuilt Docker image to speed up the startup, also clones the repository by default, so the lines are again redundant

If you agree that this is the case, we can remove them and speed up building the image.

yakutovicha: I suggest we have a quick chat about this today during the meeting, and then we can decide.

edoardob90:
Options:

### Leave the `COPY` command
- We will have a redundant copy of the entire material when we're using Codespaces: one in `/home/jovyan/python-tutorial` and another in `/workspaces/python-tutorial` (clone by default by GitHub)
- Docker image is larger and takes slightly longer to build

### Remove `COPY`
- Users must clone the repository to have the materials available inside the container
- Change the README accordingly: the bind-mount directory is not `/home/jovyan/work` but `/home/jovyan/python-tutorial`

edoardob90:
@yakutovicha I think we need to resume/finalize this. It would simplify *a lot* the build step of the image, and solve the [recent problem](https://github.com/empa-scientific-it/python-tutorial/actions/runs/14821016433/job/41614917501) with the workflow run that, all of a sudden, ran out of disk space on the hosted runner.

If we're deploying **only** on Renkulab, the `COPY` instruction to have the repository available inside the container **is not** needed. Renku already clones the repository and mounts it inside the container at runtime.

yakutovicha (approved):
Great job, grazie @edoardob90!

PR #265, opened 2024-11-26
Fix traceback issue

edoardob90:
`%%ipytest` magic will disable the default traceback only during the execution of the tests. This is done with a `traceback_handling` context manager.


edoardob90:
I added a couple of other loosely related (but very small) fixes that I noticed while going over the notebooks.

despadam (approved):
Looks alright, thanks @edoardob90 !

Issue #241, opened 2024-05-08
Errors found in the `modules and packages` notebook


despadam:

despadam:
In `exercise on installing the package` specify that they need to be in the parent `mypackage` directory. They were already there and since the exercise asks them to execute `cd mypackage` then they end up in the `mypackage` subdirectory where the `.toml` file does not exist.

yakutovicha:
> (not sure about this one) In `Exercise on importing from parent modules` should they import from `mypackage` or `mypackage,geometry`?

Importing from `mypackage` is the way to go, because `mypackage.__init__.py` already imports the `geometry` module.

PR #266, opened 2024-11-26
Address minor issues in the "modules and packages" notebook

yakutovicha:
Fixes https://github.com/empa-scientific-it/python-tutorial/issues/241

### DIFF
--- a/modules_and_packages.ipynb
+++ b/modules_and_packages.ipynb
@@ -138,7 +138,7 @@
     "Move the `Point` class and the `distance` function to a new file called `point.py`:\n",
     "\n",
     "1. Create a new file called `point.py` in the same directory as this notebook.\n",
-    "2. Copy the `Point` class and the `distance` function to the new file.\n",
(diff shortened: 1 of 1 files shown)


despadam:
Addresses #241

Snowwpanda (commented):
Sasha thanks for looking through the exercises and making these improvements.

I went through the exercises and they all work, however I think the abstaction of this part of the tutorial is very high.. which isnt bad but I wouldn't mind if you do almost all the exercises together in plenum because I doubt anyone will have any experience here anyway and it's almost impossible for them to find mistakes when they copy files into wrong folders etc.

despadam (approved):

yakutovicha:
> ... I wouldn't mind if you do almost all the exercises together in plenum because I doubt anyone will have any experience here anyway and it's almost impossible for them to find mistakes when they copy files into wrong folders etc.

I agree, I can do that 👍

Issue #81, opened 2023-04-23
Naming: enumerate notebooks


edoardob90:
It might be better to enumerate the notebooks (starting with `0-intro.ipynb`) so that they are sorted in GitHub as well

edoardob90:
I **love** that notebooks are now neatly ordered. Thanks @despadam for finalizing this one!

PR #294, opened 2025-05-02
Fix broken test module names

edoardob90:
Aftern #81 the test modules are not found. A Python module's name cannot contain `-`, so I renamed everything consistently using underscores. The order is preserved.

I also fixed a Ruff lint error from OOP tests.


edoardob90:
Need to fix the internal links between notebooks before merging.

despadam:
I never thought of that 🤦 I am sorry! I can take over this PR if you have other things to prioritize

despadam (approved):
Renaming looks good, thanks for taking care of this. I manually checked each notebook to confirm that the test modules are always being loaded correctly.

Also:
- Fixed all internal broken links, similarly to #294 
- Fixed how `testsuite` and `ipytest` were being loaded in `11_functional_programming.ipynb`

PR #297, opened 2025-05-05
Add image annotations

edoardob90:


edoardob90:
Renku seems to have problem when pulling multi-arch images.

PR #298, opened 2025-05-06
Optimize multi-arch Docker images

edoardob90:
- Reduce image size by implementing multi-stage build
- Remove torchaudio package and optimize PyTorch installation
- Add GPU/CPU variant support via build arguments
- Enable multi-architecture builds (amd64/arm64) with proper manifests
- Fix Renku compatibility by setting provenance: false
- Simplify container environment variables and working directory
- Create architecture-specific tags for better image management


yakutovicha (approved):
LGTM!

edoardob90:
Maybe we should also adjust the README? Saying that `latest` is now pulling the CPU image by default. If you want the other variant, should specify `cuda`

yakutovicha (approved):

PR #299, opened 2025-05-06
Libraries: Image classification with PyTorch

fabioacl:
I am uploading a first version of my notebook and also the Python script with the functions needed for running some pytests. The notebook is almost complete.


25_libraries_image_classification.ipynb
```diff
@@ -0,0 +1,1646 @@
...
+    "from pytorch_grad_cam.utils.model_targets import ClassifierOutputTarget\n",
+    "from pytorch_grad_cam.utils.image import show_cam_on_image"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "5",
+   "metadata": {},
+   "source": [
+    "## References\n",
```
despadam, line 121: I would move the `References` section to be before `Libraries`. Then remember to check that the table of contents stays up to date

25_libraries_image_classification.ipynb
```diff
@@ -0,0 +1,1646 @@
...
+   "id": "3",
+   "metadata": {},
+   "source": [
+    "- NumPy\n",
+    "- Matplotlib\n",
+    "- Scikit-learn\n",
+    "- OpenCV-Python\n",
+    "- PyTorch\n",
+    "- Albumentations\n",
+    "- PyTorch GradCAM"
```
despadam, line 82: For those that have an existing notebook, we can also put a link to that notebook

edoardob90:
One small thing: we should probably rename these "project" notebooks like `XX_project_name.ipynb` without the `library_`/`libraries_` prefix. Those prefixes where meaningful when the notebooks weren't in progressive order.

25_libraries_image_classification.ipynb
fabioacl: Fixed in [4a9ec7c](https://github.com/empa-scientific-it/python-tutorial/pull/299/commits/4a9ec7c3dde267f3f845dda0931817bcc7976326)

yakutovicha (approved):
Obrigado!

PR #301, opened 2025-05-07
Update OpenAI models

edoardob90:
Update to most recent/stable models that support structured outputs


despadam (approved):
Works as expected

PR #311, opened 2025-05-08
Fix errors in Functional Programming

### LINKED ISSUE #310: Errors found in `functional_programming`
The signature of the test function `test_multiples_of_n` should be `function_to_test: Callable[[List[int], int], int]`

edoardob90:
Fix #310


edoardob90:
See also #312

PR #316, opened 2025-05-08
Customize Jupyter & IPython config

### LINKED ISSUE #281: Enable editing hidden files in Jupyter
In the `modules and packages` section, people must manipulate hidden files (e.g. `.gitignore`). At the moment, Jupyter doesn't allow that. [Here is an explanation](https://jupyterlab.readthedocs.io/en/stable/user/files.html#displaying-hidden-files) of how to enable it.

The problem was first reported [here](https://github.com/empa-scientific-it/python-tutorial/issues/241#issuecomment-2100447398), but having a separate issue for that makes more sense.

### LINKED ISSUE #315: Automatic import of `typing`
We often using `typing` for type hints on reference solutions. If somebody copies that solution in a notebook, often it fails because the `typing` module is not automatically imported.

A workaround is to configure IPython to automatically import `typing` as described here: https://ipython.readthedocs.io/en/stable/interactive/tutorial.html#startup-files. We just need to create a Python script inside the folder `~/.ipython/profile_default`, which can be created via the command `ipython profile create` (but it's usually already there).

edoardob90:
Fixes #281  and fixes #315

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,7 +79,6 @@ target/
 
 # IPython
 profile_default/
-ipython_config.py
 
 # pyenv
 .python-version
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,15 +64,15 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/*
 
 # System-wide Jupyter Server config
-COPY docker/jupyter_server_config.py /etc/jupyter/
+COPY docker/jupyter_server_config.py /opt/conda/etc/jupyter/
 
 # Jupyter Lab settings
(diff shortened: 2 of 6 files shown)


yakutovicha:
@edoardob90, I will test it after the tutorial, I don't want to deal with unexpected problems in the middle of it 🥵

edoardob90:
> @edoardob90, I will test it after the tutorial, I don't want to deal with unexpected problems in the middle of it 🥵

It's not urgent. And in any case, it won't affect sessions already running on Renku

yakutovicha (approved):
Tested on Renku, seems to work fine, thanks @edoardob90 🙏

PR #323, opened 2025-05-12
Plotly intro tutorial

Snowwpanda:
Notebook 30 and all necessary files with it


edoardob90:
I think you forgot to update the dependencies in `docker/environment.yml`. Maybe we can skip it for the ongoing tutorial, but we must update that file accordingly to avoid forgetting about them. We can open a new Issue for that as a reminder.

yakutovicha (approved):
thanks, @Snowwpanda!

PR #328, opened 2025-05-12
Change GitHub workflow

edoardob90:
Run only if Dockerfile or any file related is changed. Do not automatically run if notebooks' content is updated, since that update can be easily pulled via git.


yakutovicha (approved):
LGTM!

Issue #282, opened 2024-12-05
Create an introductory notebook about syntax


edoardob90:
We know that people who are meeting Python for the first time struggle a lot with syntax. The idea is to have a introductory notebook **only about** syntax (an overview).

yakutovicha:
Thanks, @despadam, for taking this issue over.
Here’s how I would outline the “Introduction to Python” notebook.


### 1. References

Add some links from #244 so readers can see what other material is available.


### 2. “Hello world!” in Python

Start with the simplest possible program:

```python
print("Hello world!")
```

Use this to explain:

* what an **expression** is,
* what a **string** is.

(You could also mention how the interpreter executes code line-by-line, to ground readers in Python’s interactive nature.)


### 3. Basic datatypes and operations

Give a short interactive taste of numbers and strings:

```python
1 + 1
'Hello' + 'world!'
type(1)
```

Emphasise that type attribution is **implicit** and **context-dependent**.
(Optionally: show `float`, `bool`, or `list` for variety.)


### 4. Conditional execution

Introduce a simple `if-else` example to show how control flow can deviate:

```python
if 5 > 3:
    print("True branch")
else:
    print("False branch")
```


### 5. Loops

Demonstrate `while` and `for` loops and note that they allow **repeated execution**:

```python
for i in range(3):
    print(i)
```

(You could tie this to a small practical example, e.g. summing a list.)


### 6. Functions

When code should be reused, introduce functions:

```python
def greet(name):
    return f"Hello, {name}!"
```

Explain the idea of inputs, outputs, and code modularity.


### 7. Object-oriented programming

When functions are attached to objects, gives a short taste of OOP:

```python
class Greeter:
    def __init__(self, name):
        self.name = name
    def greet(self):
        return f"Hello, {self.name}!"
```

Keep it lightweight—just enough to show how data and behaviour live together.


### Side remarks

1. Each time a new topic is introduced, end with something like
   *“For more information, see [link]”*, where the link points to the relevant detailed notebook.
2. Consider adding a **coherent storyline**—for instance, reusing the small *geometry* examples from the “Modules and packages” section.
   (A consistent mini-project helps learners see how ideas build on each other.)
3. Optionally, you might end with a small “Putting it all together” cell. E.g., compute the area of a shape using a function and a class to reinforce all concepts in one place.

Issue #244, opened 2024-05-16
Add (link to) material for complete beginners


despadam:
Since some people have mentioned that they struggle with the syntax or the basic concepts of not only Python, but coding in general if they have never done it before, here are a few suggestions on how to provide additional help:
- Increase the total length of the tutorial to 4 days. The 1st day will be optional and will cover the very basics.
- Provide a knowledge test before the tutorial starts, so that we know if they need to attend the 1st day or not.
- Provide material they could read before the tutorial starts.
- Instead of writing our own material and adding an extra day, we could provide links to already existing online tutorials that teach the very basic concepts and introduce students to the syntax.

despadam:
To be decided if https://github.com/empa-scientific-it/python-tutorial/issues/282 get implemented and we create our own notebook on syntax, or if we use already existing material.

edoardob90:
This is a useful link with **many** beginner (and more) questions answered with examples: https://docs.python.org/3/faq/programming.html#programming-faq

PR #337, opened 2025-10-31
Add notebook: Introduction to Python

despadam:
Addresses: #282, #244

### DIFF
--- a/00_python_intro.ipynb
+++ b/00_python_intro.ipynb
@@ -218,7 +218,9 @@
     "```\n",
     "\n",
     "If the function needs information to work with, you put that information inside the parentheses.\n",
-    "We call this information **arguments**."
+    "We call this information **arguments**.\n",
+    "\n",
+    "We are going to dive deeper into this topic, when we talk about [Functions](./00_python_intro.ipynb#Functions) later."
    ]
   },
   {


00_python_intro.ipynb
```diff
@@ -0,0 +1,1359 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "id": "9",
+   "metadata": {},
+   "source": [
+    "### Indentation: Python's superpower\n",
+    "\n",
+    "This is one of the most important concepts in Python:\n",
+    "**Indentation** refers to the spaces (or tabs) at the beginning of a line of code.\n",
```
yakutovicha:
```suggestion
    "This is one of the most important concepts in Python: **indentation** refers to the spaces (or tabs) at the beginning of a line of code.\n",
```

00_python_intro.ipynb
```diff
@@ -0,0 +1,1359 @@
...
+   "cell_type": "markdown",
+   "id": "67",
+   "metadata": {},
+   "source": [
+    "### Parameters and Arguments\n",
+    "\n",
+    "You can see that the function we defined above is expecting to receive a value, for the variable called `name`.\n",
+    "The variable `name` itself is called a **parameter**.\n",
+    "\n",
+    "The value that we assign to this parameter, for example `\"Alice\"`, is called an **argument**.\n",
```
yakutovicha:
```suggestion
    "The value that we assign to this parameter, for example, `\"Alice\"`, is called an **argument**.\n",
```

yakutovicha (changes requested):
Great work, @despadam. Just a bunch of minor suggestions.

00_python_intro.ipynb
```diff
@@ -0,0 +1,1358 @@
...
+   "cell_type": "markdown",
+   "id": "18",
+   "metadata": {},
+   "source": [
+    "### Comments: notes to yourself\n",
+    "\n",
+    "**Comments** are lines in your code that Python ignores.\n",
+    "They are notes for humans (including your future self) to understand what the code does.\n",
+    "\n",
+    "In Python, comments start with the `#` symbol:"
```
Snowwpanda:
```suggestion
    "In Python, comments start with the `#` symbol. Everything on the same line after the `#` symbol is ignored by the computer:"
```

00_python_intro.ipynb
```diff
@@ -0,0 +1,1358 @@
...
+   "cell_type": "markdown",
+   "id": "26",
+   "metadata": {},
+   "source": [
+    "### How Python executes this code\n",
+    "\n",
+    "Python is an **interpreted** language, which means that the Python interpreter reads and executes your code line by line, from top to bottom.\n",
+    "This interactive nature makes Python great for experimentation and learning.\n",
+    "\n",
+    "You can see this in action in Jupyter notebooks like this one: each cell is executed independently when you run it."
```
Snowwpanda: I think this notebook is missing an explanation of Jupyter notebooks and this would be a good point to explain. What are cells, what is the command to run. Just the basics so that it can be tried out in the next section. (and use some variants as some may be using vs code etc.)
Maybe also why we use juypter notebooks in general and that they are widely used in scientific communities?

00_python_intro.ipynb
```diff
@@ -0,0 +1,1358 @@
...
+    "Numbers are added mathematically, while strings are joined together.\n",
+    "This is a simple example of how Python adapts behavior based on data types."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "36",
+   "metadata": {},
+   "source": [
+    "### The `type()` function\n",
```
Snowwpanda: I think the type section is a bit too much, this is handled in the next notebook and can be referenced to be handled in a later chapter, for now i think "numbers" and "strings" is enough to explain dynamic typing. 
but not a strong opinion, we can also leave it...

00_python_intro.ipynb
```diff
@@ -0,0 +1,1358 @@
...
+    "#### 3. Case sensitivity\n",
+    "\n",
+    "Python is **case-sensitive**, which means it treats uppercase and lowercase letters as different:\n",
+    "\n",
+    "```python\n",
+    "name = \"Alice\"   # This is different from...\n",
+    "Name = \"Alice\"   # ...this variable\n",
+    "```\n",
+    "\n",
+    "Python understands the variables `name` and `Name` as completely different from each other."
```
Snowwpanda: I missed an explanation of what a variable is, not quite sure where it would go, but probably somewhere around here?

Snowwpanda (changes requested):
Great content! Good work @despadam , i hope this will help many people, and we can get feedback from ppl after the first session. I have some smaller mostly optional changes.

00_python_intro.ipynb
edoardob90: During the tutorial, especially the Intro, nobody should be using VS Code. If we make sure that everybody is using Jupyter (via Renku), we remove some pain points. We could definitely explain what Jupyter is in general, but I think it goes a bit out of scope of this introduction. Jupyter is simply the most convenient interactive interface for Python.

00_index.ipynb
```diff
@@ -7,6 +7,7 @@
     "# Introduction\n",
     "\n",
     "- [Introduction to the tutorial](./00_intro.ipynb)\n",
+    "- [A quick tour of Python](./00_python_intro.ipynb)\n",
```
edoardob90:
```suggestion
    "- [Introduction to the Python programming language](./00_python_intro.ipynb)\n",
```

00_python_intro.ipynb
```diff
@@ -0,0 +1,1425 @@
...
+    "```\n",
+    "\n",
+    "The `=` symbol is called the **assignment operator**.\n",
+    "Note that it does not mean \"equals\" like in math. \n",
+    "It means \"store the value on the right into the variable on the left.\"\n",
+    "\n",
+    "Variables are useful because they allow you to:\n",
+    "- **Reuse** the same value multiple times\n",
+    "- **Change** the value later if needed\n",
+    "- Make you code more **readable** (using `name` instead of just `Alice`)"
```
edoardob90: I would add a note that "assignment" (binding names to objects) also happens with:
- Function definitions
- `import` statements
- Some language constructs (`for`, `while`, `with` all imply assignments)

We don't have to go into details here but they are worth a mention.

edoardob90 (changes requested):
Very much complete: thorough but not "too much" for complete beginners. Thanks @despadam! I've only two quite small suggestions.

edoardob90 (dismissed):
LGTM 🚀

fabioacl (commented):
Only some minor details and it is good to go. Very good job @despadam. I really liked the last example where you use all the concepts you introduced before.

yakutovicha (approved):
Great job, @despadam!

PR #339, opened 2025-11-07
Add instructions on using Renku

### LINKED ISSUE #289: Add instructions on how to use Renku
Add a section to `intro.ipynb` with an explanation on how to use Renku

edoardob90:
Fix #289


00_intro.ipynb
```diff
@@ -103,6 +104,72 @@
...
+    "\n",
+    "### Start a New Session\n",
+    "\n",
+    "After you created your account, go back to the [tutorial GitHub's page](https://github.com/empa-scientific-it/python-tutorial) and click on the Renku badge to start a new session:\n",
+    "\n",
+    "<img src=\"images/start_renku_session.png\" alt=\"Start a new Renku session\" width=\"800\"/>\n",
+    "\n",
+    "### Working in a Session\n",
+    "\n",
+    "Once your session starts, you'll be in a **JupyterLab environment** where you can work with notebooks, Python files, and terminal commands.\n",
```
despadam: Oftentimes multiple sessions are created, for example a Python environment, a Jupyter one, etc. Should we specify which one they should use?

edoardob90: I already deleted all the temporary/test sessions for the upcoming tutorial. There're only two left, and I renamed the "default" to "Python Tutorial Launcher". I'll add it to the notebook for reference, but it's better if we just tell them directly.

PR #346, opened 2025-11-12
Fix modules and packages notebook

### LINKED ISSUE #280: Errors found in `modules_and_packages`
- [ ] Generating the github token is not documented so well, we need to tell them which access rights they need to use and maybe more

yakutovicha:
fixes #280


06_modules_and_packages.ipynb
```diff
@@ -878,7 +878,22 @@
...
     "For example, John Doe can use `johndoe`, `john.doe`, `johnd`, `jdoe`, etc.\n",
     "\n",
-    "2. [Create an access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token#creating-a-personal-access-token-classic) for your account that will be used instead of your password."
+    "2. Enable passwordless connection to your GitHub profile.\n",
+    "\n",
+    "   1. For that, you first need to open the Terminal and type the following:\n",
+    "       ```bash\n",
+    "       $ ssh-keygen -t rsa\n",
+    "       ```\n",
+    "       Then hit `Enter` three times to accept the default values.\n",
```
edoardob90: If you changed the `ssh-keygen` command, you need to confirm the no-passphrase twice:

```suggestion
    "       Then hit `Enter` two times to accept the default values.\n",
```

06_modules_and_packages.ipynb
```diff
@@ -927,6 +943,18 @@
...
+    "<div class=\"alert alert-block alert-warning\"><b>Warning:</b>\n",
+    "    Here, Git might complain that it doesn't know who you are.\n",
+    "    Fix that by providing your email and name specified during the GitHub registration.\n",
+    "    <br>\n",
+    "    <code> git config --global user.email \"you@example.com\" </code>\n",
+    "    <br>\n",
+    "    <code> git config --global user.name \"Your Name\" </code>\n",
+    "    <br>\n",
+    "    After this is done, repeat the <code>git commit ...</code> command from above.\n",
+    "</div>\n",
```
edoardob90: Git has been never set up in the Renku session. We can assume these steps must be done *before* you add or commit. I would move them after `git init`, and make them non-optional. Also, it doesn't harm to type them twice with the same username/email.

edoardob90 (changes requested):
Thanks @yakutovicha! SSH flow works.

Maybe a few suggestions to make it even simpler. Plus a couple of general comments.

06_modules_and_packages.ipynb
edoardob90: Linked to the above.

06_modules_and_packages.ipynb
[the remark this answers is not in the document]
yakutovicha: done in https://github.com/empa-scientific-it/python-tutorial/pull/346/commits/bcddbb7dad331d04a69b82547d239842ee9b98b4

edoardob90 (approved):
Looks great and ready! 💯 If @Snowwpanda wants to make other suggestions... Otherwise we can merge.

